PR #56, opened 2020-03-12
[ToDo] Add action code

ssss141414:
### Purpose

### Changes

### Tests

### Feature Plan

#### General


skills/csharp/todoskill/wwwroot/manifest/manifest-1.0.json
```diff
@@ -118,11 +118,20 @@
         }
       }
     },
-    "todoList": {
-      "type": "array",
-      "description": "List of todo tasks",
-      "items": {
-        "type": "string"
+    "todoListInfo": {
```
bobokids: we may only use one ActionResult like todolistinfo? for DeleteTodo, the list is just null.

skills/csharp/tests/todoskill.tests/Flow/ToDoSkillTestBase.cs
```diff
@@ -183,5 +185,28 @@ protected string[] ActionEndMessage()
...
         }
+
+        protected Action<IActivity> CheckForEoC(int taskCount)
+        {
+            return activity =>
+            {
+                var eoc = (Activity)activity;
+                Assert.AreEqual(ActivityTypes.EndOfConversation, eoc.Type);
+
+                if (eoc.Value is ActionResult)
```
xieofxie: If for testing, should assert the type of eoc.Value? Or if return a wrong type, we don't know

skills/csharp/tests/todoskill.tests/Flow/Utterances/AddToDoFlowTestUtterances.cs
```diff
@@ -74,6 +77,19 @@ public AddToDoFlowTestUtterances()
 
         public static string AddTaskWithContentAndCustomizeListType { get; } = "add history to my homework list";
 
+        public static string AddToDoActionName { get; } = "AddToDo";
```
xieofxie: In my code, I unify these names together with the switch in RouteStep

skills/csharp/todoskill/Dialogs/MainDialog.cs
```diff
@@ -385,5 +443,34 @@ private void InitializeConfig(ToDoSkillState state)
                 }
             }
         }
+
+        private async Task DigestActionInput(DialogContext dc, object actionInput)
```
xieofxie: If only have two like ToDoInfo, ListInfo, could use two functions to specify the class to use.
Or use a base class / interface to unify the function.

For current code, it will always end in first try because json will just fill the attributes

skills/csharp/todoskill/wwwroot/manifest/manifest-1.1.json
```diff
@@ -183,11 +183,20 @@
...
-      "type": "array",
-      "description": "List of todo tasks",
-      "items": {
-        "type": "string"
+    "todoListInfo": {
+      "type": "object",
+      "properties": {
+        "actionSuccess": {
+          "type": "bool",
+          "description": "result of action"
```
xieofxie: Result instead of result

skills/csharp/tests/todoskill.tests/Flow/Utterances/AddToDoFlowTestUtterances.cs
ssss141414: Nice design!

skills/csharp/todoskill/Dialogs/MainDialog.cs
ssss141414: Thanks. I delete the second part, because the first part can fill all properties already.

xieofxie: No, you misunderstand the mechanism of json. In this way, use could use every action with ToDoInfo and fill all state variables. It may be not harm this time but it is a bad design.
Anyway, pay attention next time.

skills/csharp/tests/todoskill.tests/Flow/ToDoSkillTestBase.cs
xieofxie: User should indicate the type of eoc.Value when calling CheckForEoC like CheckForEoC(Enum valueType, int taskCount)

For current code, if one flow returns ActionResult while actually we expect TodoListInfo, it is not checked.

ssss141414: Sure. 
Is Enum needed here? Or we just use CheckForEoC(Type valueType, int taskCount)? Because we only have two types...

xieofxie: No, it is just an example. Any kind of indicator is fine

PR #76, opened 2020-03-13
[Event] SDK & LG migration

bobokids:
### Purpose
Migrate event skill to 4.8.0

### Changes
. SDK migration
. LG support

### Tests

### Feature Plan

#### General


skills/csharp/experimental/eventskill/Responses/ResponsesAndTexts/ResponsesAndTexts.lg
```diff
@@ -0,0 +1,3 @@
+[import](..\FindEvents\FindEventsResponses.lg)
```
xhr0804: why "FindEventResponse" not exist in other language file?

xieofxie: Because there is only FindEventsResponses.lg and no other locales..

xhr0804: ok..

skills/csharp/experimental/eventskill/Startup.cs
```diff
@@ -103,11 +97,7 @@ public void ConfigureServices(IServiceCollection services)
             services.AddHostedService<QueuedHostedService>();
 
             // Configure responses
-            services.AddSingleton(sp => new ResponseManager(
-                settings.CognitiveModels.Select(l => l.Key).ToArray(),
-                new MainResponses(),
-                new SharedResponses(),
-                new FindEventsResponses()));
+            services.AddSingleton(LocaleTemplateManagerWrapper.CreateLocaleTemplateManager("en-us"));
```
xieofxie: Actually it has 6 locales except for FindEventsResponses

skills/csharp/experimental/eventskill/Responses/Shared/SharedResponses.lg
```diff
@@ -0,0 +1,57 @@
...
+    Text = ${ActionEnded.Text(Data)}
+    Speak = ${ActionEnded.Text(Data)}
+    Attachments = ${if(Cards == null, null, foreach(Cards, Card, CreateCard(Card)))}
+    AttachmentLayout = ${if(Layout == null, 'list', Layout)}
+    InputHint = acceptingInput
+]
+
+# ErrorMessage(Data, Cards, Layout)
+[Activity
+    Text = ${ErrorMessage.Text(Data)}
```
KayMKM: this lg is different with other language.

skills/csharp/experimental/eventskill/Responses/Shared/SharedResponses.zh-cn.lg
```diff
@@ -0,0 +1,58 @@
...
+    Text = ${NoAuth.Text(Data)}
+    Speak = ${NoAuth.Text(Data)}
+    Attachments = ${if(Cards == null, null, foreach(Cards, Card, CreateCard(Card)))}
+    AttachmentLayout = ${if(Layout == null, 'list', Layout)}
+    InputHint = expectingInput
+]
+
+# AuthFailed(Data, Cards, Layout)
+[Activity
+    Text = ${AuthFailed.Text(Data)}
```
KayMKM: this lg is different with other language.

skills/csharp/experimental/eventskill/Startup.cs
bobokids: fixed.

skills/csharp/experimental/eventskill/Responses/ResponsesAndTexts/ResponsesAndTexts.lg
bobokids: leave as it is.

skills/csharp/experimental/eventskill/Responses/Shared/SharedResponses.zh-cn.lg
bobokids: Good catch. changed other languages to zh-cn.lg

skills/csharp/experimental/eventskill/Responses/ResponsesAndTexts/ResponsesAndTexts.lg
lzc850612: so what happens when user is in a different locale that doesn't have FindEventsResponses.{that locale}.lg imported, there would be exceptions right? since the code still looks for the same template. should probably do some easy translation with Google translate for those other languages.

bobokids: The skill actually does not work due to the API issue (API not work now). it's also dropped from the experimental skill list in the skill document. So I just updated the skill to align with the SDK change. The response message for FindEvent dialog has the multi-lang support issue but I think we can leave it there until we have further plan to support this skill.

PR #88, opened 2020-03-17
[Weather] Migrate weather API to Azure Maps

ssss141414:
### Purpose
Migrate weather API to Azure Maps
https://github.com/microsoft/botframework-skills/issues/29

### Changes

### Tests

### Feature Plan

#### General


lzc850612:
are we just completely getting rid of AccuWeather now? can we define interface and use both service provider (AccuWeather and Azure Maps) with Azure maps being the default one?

ssss141414:
> are we just completely getting rid of AccuWeather now? can we define interface and use both service provider (AccuWeather and Azure Maps) with Azure maps being the default one?

We have discussed with this problem, and we decided to only keep one service.

skills/csharp/tests/weatherSkill.tests/Flow/ForecastFlowTests.cs
```diff
@@ -0,0 +1,72 @@
...
+                .Send(ForecastUtterances.WeatherForecastActionWithCoordinates)
+                .AssertReply(CheckForEoC())
+                .StartTestAsync();
+        }
+
+        [TestMethod]
+        public async Task Test_AskWeatherWithLocation()
+        {
+            await this.GetTestFlow()
+                .Send(string.Empty)
```
xieofxie: use SendConversationUpdate()

PR #127, opened 2020-03-22
[ToDo] Fix issue

ssss141414:
### Purpose
Fix an issue: When first time signing in, cause an error.
https://github.com/microsoft/botframework-skills/issues/128

### Changes

### Tests

### Feature Plan

#### General


ssss141414:
> Could this be reproduced in test? If so, please add a test to verify the fix. Thanks!

It's seems that our skill test flows don't have those oauth steps? So maybe it can't be tested.

skills/csharp/todoskill/Dialogs/ShowToDoItemDialog.cs
```diff
@@ -139,8 +139,21 @@ public async Task<DialogTurnResult> ShowTasks(WaterfallStepContext sc, Cancellat
 
                 state.ListType = state.ListType ?? ToDoStrings.ToDo;
                 state.LastListType = state.ListType;
+
+                ToDoLuis.Intent topIntent = ToDoLuis.Intent.ShowToDo;
                 var luisResult = sc.Context.TurnState.Get<ToDoLuis>(StateProperties.ToDoLuisResultKey);
```
xieofxie: I have create a tested here https://github.com/microsoft/botframework-skills/commit/c6db57822c2ecbde2db033a0c2266834d428da98

Because the token is set when creating TestAdapter, so there is no login prompt.
I manually log out in this test to trigger login prompt.

And I found another bug: should not rely on ToDoLuis here. Because when token is set by number like in VA, webchat, ToDoLuis will be None.

ssss141414: Thanks for your test! Can you create a PR to add it?

Af for the bug, I think it's OK here. PVA also send a number to complete Oauth step and I have tested it. I set ShowToDo as default intent. If ToDoLuis is null, intent will not be changed, and the following logic goes well.
Please let me know if I misunderstand this bug.

xieofxie: No, the bug should be fixed. In current logic, ListType will not be set and the test I added will fail.
You could check out the branch to see it https://github.com/microsoft/botframework-skills/compare/hualxie/todo_fix

It shares same flow with Test_ShowToDoItems but it won't work

ssss141414: Oh, I found the bug. Because PVA's ToDoLuis is null, but this time ToDoLuis is with a None Intent. So this code works in PVA but not here.
So I modify If condition here:
    if (luisResult != null && luisResult.TopIntent().intent != ToDoLuis.Intent.None)
I also filter None Intent here. This makes sense, because in ToDo skill, None intent appears here only when sign in.
Thanks a lot!

xieofxie: It will be a little weird if it is triggered by a ShowNextPage, ShowPreviousPage, but it is fine.

PR #149, opened 2020-03-30
update skills to use oauthcredentials

lzc850612:
### Purpose

Add OAuthCredentials in appsettings for productivity skills to enable 'SSO' among skills
Add code to take in OAuthCredentials when creating MultiProviderAuthDialog
Update solutions lib to 1.0.0

### Changes

### Tests

### Feature Plan

#### General


xieofxie:
So oauthCredentials should be set same as VA's?

lzc850612:
> So oauthCredentials should be set same as VA's?

yeah, or whichever resource that contains the oauth connection settings. this way if multiple skills are using the connections from the same bot, then user only has to login once.

PR #156, opened 2020-04-03
Rideshare Skill Stub

peterinnesmsft:
### Purpose
Adding a stub Rideshare skill and test project to start development of the experimental rideshare skill.

### Changes
Two new projects, RideshareSkill and RideshareSkill.Tests, have been added. These are as-is fresh from installing via the Skill Template VSIX.

### Tests
Yes - covered by stub tests installed from the VSIX.

### Feature Plan
N/A

#### General


lzc850612:
same as the other PR. if this is not against master, i'm not sure we should have people look at it because when you have to merge it to master people have to look at it again and it's very easy to miss things, if that makes sense.

peterinnesmsft:
> same as the other PR. if this is not against master, i'm not sure we should have people look at it because when you have to merge it to master people have to look at it again and it's very easy to miss things, if that makes sense.

I'm going to challenge that. If the final merge to master ends up being small, then I could buy it - but I've seen too many cases where people end up building giant branches to be merged back to master. I don't think it's fair on reviewers to have to look at ginormous reviews, and can also lead to misses because less attention was paid in reviewing the changes, particularly if there is pressure to get checked in. I am much more of a fan of little and often.

Let me know if you disagree.

ryanlengel:
The PR template has a section called **Feature Plan** that has gone largely unused but I think that there are valid cases for it.

* Child PRs mention that they are part of a larger feature
* Parent PR references the Child PRs to show that code reviews for the individual components have already occurred
* Making changes in a Sample & the SDK that require new packages published to pick up on said changes... _(just another example)_


I'm confused as to whether the `peinnes/rideshare_master` branch is going to be merged into next or master?

lzc850612:
> > same as the other PR. if this is not against master, i'm not sure we should have people look at it because when you have to merge it to master people have to look at it again and it's very easy to miss things, if that makes sense.
> 
> I'm going to challenge that. If the final merge to master ends up being small, then I could buy it - but I've seen too many cases where people end up building giant branches to be merged back to master. I don't think it's fair on reviewers to have to look at ginormous reviews, and can also lead to misses because less attention was paid in reviewing the changes, particularly if there is pressure to get checked in. I am much more of a fan of little and often.
> 
> Let me know if you disagree.

Thanks Peter for the comment. 

I disagree. The most important thing for PRs (I believe) is that every single PR needs to be guarded by certain checks (e.g. azure devops build). If we don't have the checks, there's no guarantee the PRs are good even though they are approved. (you can argue that the checks won't guarantee things either, but it's the next best thing and we're always improving it with test coverage etc). if we create PRs against a branch that doesn't have checks (private branches), these things will/might happen:

1. there's no checks running for them so there's no guarantee the changes are good
2. it's easy to slip commits through by simply 'git push'. i know people normally wouldn't do it, but if it's a public repo, we can't be sure
3. when you finally merge the big change into master/next branch that has check running, people wouldn't actually look at that PR again since they'll assume they've looked at all that's included, but there could easily be commits by #2 

Even though the alternative is bigger PRs, i think we can all agree that the three things above are worse problems. 

And IMHO bigger PRs are not always a problem, nor are they a problem we can't solve. i think we should always be able to divide big PRs into smaller atomic ones that make sense and work themselves (e.g. in phases). in most cases a better design needs to be done if that's not possible. and if not the PR owners can always add initial comments here and there in the PR to point people to different places in the PR to help them to look at a big one.  

let me know what you think and feel free to sync up offline more about this.

PR #169, opened 2020-04-15
[Calendar] Accept-Delete-Update and BookMeetingRoom

xhr0804:
### Purpose
An initial version for Accept/Delete/Update meeting and BookMeetingRoom.
### Changes

### Tests

### Feature Plan

#### General

### DIFF
--- a/skills/declarative/calendarskill/dialogs/AcceptMeetingDialog/AcceptMeetingDialog.dialog
+++ b/skills/declarative/calendarskill/dialogs/AcceptMeetingDialog/AcceptMeetingDialog.dialog
@@ -1,0 +1,155 @@
+{
+  "$kind": "Microsoft.AdaptiveDialog",
+  "$designer": {
+    "id": "G4-9JT",
+    "name": "AcceptMeetingDialog"
+  },
+  "autoEndDialog": "true",
+  "defaultResultProperty": "dialog.result",
+  "triggers": [
+    {
+      "$kind": "Microsoft.OnBeginDialog",
+      "$designer": {
+        "name": "BeginDialog",
+        "id": "4-oY4n"
+      },
+      "actions": [
+        {
+          "$kind": "Microsoft.BeginDialog",
+          "$designer": {
+            "id": "U8m0n_",
+            "name": "search a meeting"
+          },
+          "dialog": "SearchMeetings",
+          "resultProperty": "dialog.meeting"
+        },
+        {
+          "$kind": "Microsoft.IfCondition",
+          "$designer": {
+            "id": "yLVPWu"
+          },
+          "condition": "dialog.meeting == null || count(dialog.meeting) == 0",
+          "actions": [
+            {
+              "$kind": "Microsoft.EndDialog",
+              "$designer": {
+                "id": "FJ1TMh"
+              }
+            }
+          ]
+        },
+        {
+          "$kind": "Microsoft.ConfirmInput",
+          "$designer": {
+            "id": "w8sm4A"
+          },
+          "maxTurnCount": 3,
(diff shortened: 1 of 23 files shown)


xieofxie:
I think you could just discuss and test with @KayMKM and then merge this.
We don't have tests and we are in the process of fast try-and-implement, reviews are somehow difficult and slow

skills/declarative/calendarskill/dialogs/AcceptMeetingDialog/AcceptMeetingDialog.dialog
```diff
@@ -0,0 +1,155 @@
...
+      ]
+    },
+    {
+      "$kind": "Microsoft.OnCustomEvent",
+      "$designer": {
+        "id": "JVVCde"
+      },
+      "actions": [
+        {
+          "$kind": "Microsoft.OAuthInput",
```
xieofxie, line 92: one tip: use only one oauth dialog and begin them when needed. Then you don't need to set text, connection etc.

skills/declarative/calendarskill/dialogs/AcceptMeetingDialog/AcceptMeetingDialog.dialog
```diff
@@ -0,0 +1,155 @@
...
+          "text": "please sign in",
+          "title": "login",
+          "connectionName": "=settings.ConnectionName",
+          "timeout": 900000,
+          "allowInterruptions": false,
+          "alwaysPrompt": false,
+          "disabled": false
+        },
+        {
+          "$kind": "Microsoft.HttpRequest",
```
xieofxie, line 106: the same applies for httprequest. headers, type part could be shared. And all you need to fill are url, method & body

skills/declarative/calendarskill/dialogs/AcceptMeetingDialog/AcceptMeetingDialog.dialog
xhr0804: Sure, this is just to be consistent with the current version. Yue and I will have more discussion on it.

skills/declarative/calendarskill/dialogs/AcceptMeetingDialog/AcceptMeetingDialog.dialog
```diff
@@ -0,0 +1,155 @@
...
+          "$designer": {
+            "id": "cn9s6-"
+          },
+          "responseType": "Json",
+          "method": "POST",
+          "url": "https://graph.microsoft.com/beta/me/events/${dialog.meeting.id}/accept",
```
KayMKM, line 112: why use beta api?

skills/declarative/calendarskill/dialogs/AcceptMeetingDialog/AcceptMeetingDialog.dialog
xhr0804: Fot the http request, I don't think it is necessary to apply more encapsulation for current scenario. The request headers are accually not the same for the msgraph/azuresearch. The method get/post/patch/delete are also different. And we have just a few places to call http request. What's more, I think the request is clear enough?
But using a dialog to handle a http request is a choice, we may have more discussion later

skills/declarative/calendarskill/dialogs/AcceptMeetingDialog/AcceptMeetingDialog.dialog
xhr0804: I used beta api just because Place API for book meeting room is just supported by beta api. But it is not necessary for other action, I will change them.

PR #172, opened 2020-04-17
[calendar] change oauth and httprequest to dialog

xhr0804:
### Purpose
change oauth and http request to dialog to optimize the flow.
### Changes

### Tests

### Feature Plan

#### General

### DIFF
--- a/skills/declarative/calendarskill/dialogs/AcceptMeetingRequest/AcceptMeetingRequest.dialog
+++ b/skills/declarative/calendarskill/dialogs/AcceptMeetingRequest/AcceptMeetingRequest.dialog
@@ -1,0 +1,79 @@
+{
+  "$kind": "Microsoft.AdaptiveDialog",
+  "$designer": {
+    "id": "jhrSvA",
+    "name": "AcceptMeetingRequest"
+  },
+  "autoEndDialog": "true",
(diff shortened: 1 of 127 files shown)


skills/declarative/calendarskill/dialogs/AcceptMeetingRequest/AcceptMeetingRequest.dialog
```diff
@@ -0,0 +1,79 @@
...
+          "method": "POST",
+          "url": "https://graph.microsoft.com/v1.0/me/events/${dialog.meetingId}/accept",
+          "headers": {
+            "authorization": "${turn.token}"
+          },
+          "resultProperty": "dialog.api_result",
+          "contentType": "application/json"
+        },
+        {
+          "$kind": "Microsoft.IfCondition",
```
KayMKM, line 45: I think the request dialog should return the result. The following logic should be handled in the parent dialog.

xhr0804: Yes, it should return a result to make flow more robust. But I think it can just return the required data but not all the api result? For example True/False for delete/accept, MeetingRoomList for findmeetingroom.
We can discuss more offline.

PR #179, opened 2020-04-22
[ITSM] add fallback to knowledge search and two states

xieofxie:
### Purpose

Add an option fallbackToKnowledge. When true, will search knowledge for none intent:
image

Add two states active/inactive for show tickets:
image

### Changes

### Tests

### Feature Plan

#### General


skills/csharp/experimental/itsmskill/Dialogs/MainDialog.cs
```diff
@@ -230,9 +232,15 @@ private async Task<DialogTurnResult> RouteStepAsync(WaterfallStepContext stepCon
                 if (luisService != null)
                 {
                     var result = stepContext.Context.TurnState.Get<ITSMLuis>(StateProperties.ITSMLuisResult);
-                    var intent = result?.TopIntent().intent;
+                    var (intent, score) = result.TopIntent();
```
xhr0804, line 235: Do we need to check "result == null" here?

xieofxie: No, if luisService is not null, ITSMLuis will be set in OnBeginDialogAsync or OnContinueDialogAsync.

PR #231, opened 2020-05-27
Implement Implicit User Logout on Unauthorized error

sidmore:
### Purpose
The change enables logging out user when ServiceNow throws 401, this enables to get user out of the loop by implicitly logging him out.

### Changes

### Tests
Yes it is covered

### Feature Plan

#### General

### DIFF
--- a/skills/csharp/experimental/itsmskill/Dialogs/SkillDialogBase.cs
+++ b/skills/csharp/experimental/itsmskill/Dialogs/SkillDialogBase.cs
@@ -910,6 +910,36 @@ namespace ITSMSkill.Dialogs
             }
         }
 
+        protected async Task<DialogTurnResult> HandleAPIUnauthorizedError(WaterfallStepContext sc, TicketsResult ticketsResult, CancellationToken cancellationToken)
+        {
+            // Check if the error is UnAuthorized
+            if (ticketsResult.Reason.Equals("Unauthorized"))
+            {
+                // Logout User
+                var botAdapter = (BotFrameworkAdapter)sc.Context.Adapter;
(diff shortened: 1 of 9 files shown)


skills/csharp/experimental/itsmskill/Dialogs/SkillDialogBase.cs
```diff
@@ -910,6 +910,36 @@ protected async Task<DialogTurnResult> ShowKnowledgeAsync(WaterfallStepContext s
...
         }
 
+        protected async Task<DialogTurnResult> HandleAPIUnauthorizedError(WaterfallStepContext sc, TicketsResult ticketsResult, CancellationToken cancellationToken)
+        {
+            // Check if the error is UnAuthorized
+            if (ticketsResult.Reason.Equals("Unauthorized"))
+            {
+                // Logout User
+                var botAdapter = (BotFrameworkAdapter)sc.Context.Adapter;
+                await botAdapter.SignOutUserAsync(sc.Context, Settings.OAuthConnections.FirstOrDefault().Name, null, cancellationToken);
```
xieofxie, line 920: Should we sign out all? See https://github.com/microsoft/botframework-skills/blob/d6d9e9baa58c4e2ee6dbe93ca3e22d3876ba48e7/skills/csharp/experimental/itsmskill/Dialogs/MainDialog.cs#L384

darrenj: Good spot - however that "logout" logic is for a user all up, i.e. if they have multiple tokens and is really designed for something like a VA where they want to signout everything. In this case we just want to signout the ITSM connection so this seems fine.

PR #301, opened 2020-08-19
[Calendar] fix displaying wrong timezone in outlook

KayMKM:
### Purpose
fix and close:
https://github.com/microsoft/botframework-skills/issues/295

### Changes
as the design before, all api calls use utc timezone. removed the utc timezone check and set user timezone when create/update meeting. keep the design of when get time from api calls it always returns utc time.

### Tests

### Feature Plan

#### General

### DIFF
--- a/skills/csharp/calendarskill/Dialogs/CreateEventDialog.cs
+++ b/skills/csharp/calendarskill/Dialogs/CreateEventDialog.cs
@@ -612,14 +612,15 @@ namespace CalendarSkill.Dialogs
                     });
                 }
 
+                var userTimezone = state.GetUserTimeZone();
                 var newEvent = new EventModel(source)
                 {
                     Title = state.MeetingInfo.Title,
                     Content = state.MeetingInfo.Content,
                     Attendees = state.MeetingInfo.ContactInfor.Contacts,
-                    StartTime = TimeConverter.ConvertUtcToUserTime((DateTime)state.MeetingInfo.StartDateTime, state.GetUserTimeZone()),
-                    EndTime = TimeConverter.ConvertUtcToUserTime((DateTime)state.MeetingInfo.EndDateTime, state.GetUserTimeZone()),
-                    TimeZone = state.GetUserTimeZone(),
+                    StartTime = TimeConverter.ConvertUtcToUserTime((DateTime)state.MeetingInfo.StartDateTime, userTimezone),
+                    EndTime = TimeConverter.ConvertUtcToUserTime((DateTime)state.MeetingInfo.EndDateTime, userTimezone),
+                    TimeZone = userTimezone,
                     Location = state.MeetingInfo.MeetingRoom == null ? state.MeetingInfo.Location : null,
                     IsOnlineMeeting = true
                 };
--- a/skills/csharp/calendarskill/Dialogs/CreateEventDialog.cs
+++ b/skills/csharp/calendarskill/Dialogs/CreateEventDialog.cs
@@ -617,9 +617,9 @@ namespace CalendarSkill.Dialogs
                     Title = state.MeetingInfo.Title,
                     Content = state.MeetingInfo.Content,
                     Attendees = state.MeetingInfo.ContactInfor.Contacts,
-                    StartTime = (DateTime)state.MeetingInfo.StartDateTime,
-                    EndTime = (DateTime)state.MeetingInfo.EndDateTime,
-                    TimeZone = TimeZoneInfo.Utc,
+                    StartTime = TimeConverter.ConvertUtcToUserTime((DateTime)state.MeetingInfo.StartDateTime, state.GetUserTimeZone()),
+                    EndTime = TimeConverter.ConvertUtcToUserTime((DateTime)state.MeetingInfo.EndDateTime, state.GetUserTimeZone()),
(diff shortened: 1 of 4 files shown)


skills/csharp/calendarskill/Dialogs/CreateEventDialog.cs
```diff
@@ -617,17 +617,18 @@ private async Task<DialogTurnResult> CreateEventAsync(WaterfallStepContext sc, C
                     Title = state.MeetingInfo.Title,
                     Content = state.MeetingInfo.Content,
                     Attendees = state.MeetingInfo.ContactInfor.Contacts,
-                    StartTime = (DateTime)state.MeetingInfo.StartDateTime,
-                    EndTime = (DateTime)state.MeetingInfo.EndDateTime,
-                    TimeZone = TimeZoneInfo.Utc,
+                    StartTime = TimeConverter.ConvertUtcToUserTime((DateTime)state.MeetingInfo.StartDateTime, state.GetUserTimeZone()),
+                    EndTime = TimeConverter.ConvertUtcToUserTime((DateTime)state.MeetingInfo.EndDateTime, state.GetUserTimeZone()),
+                    TimeZone = state.GetUserTimeZone(),
```
peterinnesmsft, line 622: >state.GetUserTimeZone() [](start = 31, length = 23)

Rather than repeat this three times, compute the time zone once before creating the newEvent object and reuse.

skills/csharp/calendarskill/Models/EventModel.cs
```diff
@@ -392,15 +394,15 @@ public DateTime StartTime
                             msftEventData.Start = new Microsoft.Graph.DateTimeTimeZone();
                         }
 
-                        msftEventData.Start.DateTime = value.ToString("o");
+                        msftEventData.Start.DateTime = value.ToString("yyyy-MM-ddTHH:mm:ss");
```
peterinnesmsft, line 397: >yyyy-MM-ddTHH:mm:ss [](start = 71, length = 19)

What's the purpose behind changing the format here? Generally speaking, it is sensible to stick to ISO-8601 formatting for date time strings.

skills/csharp/calendarskill/Models/EventModel.cs
```diff
@@ -369,7 +369,14 @@ public DateTime StartTime
                             msftEventData.Start.DateTime = msftEventData.Start.DateTime + "Z";
                         }
 
-                        return DateTime.Parse(msftEventData.Start.DateTime).ToUniversalTime();
+                        var start = new DateTime(DateTime.Parse(msftEventData.Start.DateTime).Ticks, DateTimeKind.Unspecified);
+
+                        if (this.TimeZone.Id != "UTC")
```
peterinnesmsft, line 374: >"UTC" [](start = 48, length = 5)

Please define this as a const

peterinnesmsft:
I would expect to see some updated unit tests to reflect this change. Can we make sure we have unit test coverage for this?

skills/csharp/calendarskill/Dialogs/CreateEventDialog.cs
```diff
@@ -617,17 +617,18 @@ private async Task<DialogTurnResult> CreateEventAsync(WaterfallStepContext sc, C
                     Title = state.MeetingInfo.Title,
                     Content = state.MeetingInfo.Content,
                     Attendees = state.MeetingInfo.ContactInfor.Contacts,
-                    StartTime = (DateTime)state.MeetingInfo.StartDateTime,
-                    EndTime = (DateTime)state.MeetingInfo.EndDateTime,
-                    TimeZone = TimeZoneInfo.Utc,
+                    StartTime = TimeConverter.ConvertUtcToUserTime((DateTime)state.MeetingInfo.StartDateTime, state.GetUserTimeZone()),
```
peterinnesmsft, line 620: >TimeConverter.ConvertUtcToUserTime((DateTime)state.MeetingInfo.StartDateTime, state.GetUserTimeZone()), [](start = 32, length = 103)

This feels strange to me. Shouldn't the point of storing the Time Zone separately mean that we should just get a DateTime that we can take to be in the same time zone as specified in the state?

skills/csharp/calendarskill/Models/EventModel.cs
KayMKM: The value is a UTC time. After tostring("o"), it will be convert to a string with format like "yyyy-MM-ddTHH:mm:ssZ". The msgraph api will treat the time as an UTC time and ignore the timezone we set.

skills/csharp/calendarskill/Dialogs/CreateEventDialog.cs
KayMKM: Yes. But according to the previous design, the StartDateTime and EndDateTime should be UTC time and be used in many places. If we change it to specified timezone, we need to change a lot of code. I am not sure should we take so many changes on this bug. So I choose a simpler way to fix it.

KayMKM:
> I would expect to see some updated unit tests to reflect this change. Can we make sure we have unit test coverage for this?

This change will only effect the timezone display in outlook client. The flows will not be changed. I think current flow UT has covered these flows.

Issue #297, opened 2020-08-18
Look up event(s): Day overview


lauren-mills:
## User Story
**As an** employee of an enterprise 
**I want** to as the skill about my day ahead
**so that** I can quickly see what events are coming up.

## Description
This feature should enable the user to provide natural language queries to access their events from the MS Graph API for a given date, date range, time, time range, or datetime range.

## Conversation Design
[Transcript](https://www.figma.com/file/xsSfVmzSe5CyFlcLovplS9/Calendar-Skill?node-id=26%3A1)

[Card design and considerations](https://www.figma.com/file/xsSfVmzSe5CyFlcLovplS9/Calendar-Skill?node-id=180%3A4)

## Language Understanding
| Intent | Sample Utterances | Prebuilt Entities | Custom Entities | Patterns |
| ------ | ----------------------| ----------------- | ----------------- | --------- |
| GetEventsByDate | What's on my calendar today? <br> Read me my calendar <br> Do I have an event tomorrow? <br> What's on my agenda this weekend? <br> List all events for August 20th | DatetimeV2 | - | - |

#### GetEventsByDate
* If no datetimev2 entity provided in initial utterance, default to today’s date
* If datetimev2 entity without date is provided (i.e. time or time range) default to today’s date at that time/time range.
* If datetimev2 entity is provided with only duration (i.e. 1 hour) default to today’s date.
* Call SearchEvents dialog to execute the Microsoft Graph API call and provide dialog.datetime in dialog options.
* Call SortEvents action on the results to prepare data for the adaptive card.

#### GetEventsByResponse
* If no datetimev2 entity provided in initial utterance, default to today’s date
* If datetimev2 entity without date is provided (i.e. time or time range) default to today’s date at that time/time range.
* If datetimev2 entity is provided with only duration (i.e. 1 hour) default to today’s date.
* Call SearchEvents dialog to execute the Microsoft Graph API call and provide dialog.datetime and dialog.responseType in dialog options.
* Call SortEvents action on the results to prepare data for the adaptive card.

#### SearchEvents
* Collect parameters from dialog options (datetime, responseType).
* Call MS Graph – Search Events action with provided parameters (datetime, responseType, token, time zone)
* Return results of API call in EndDialog action

## Custom Actions
| Name | Input | Output | Notes |
| ------ | ------ | -------- | ------- |
| ParseTimex | LUIS datetimeV2 timex array | Collection of parsed results | |
[...]

PR #306, opened 2020-09-05
[Do not merge] Calendar skill v2

lauren-mills:
#297

### Purpose
This PR includes the declarative files and custom actions for the Look up event(s): Day overview feature. It also includes the ResolveDateTime action and dialog.

### Changes
- The SearchEvents dialog is pulled out separately from the GetEventsByDateTime dialog because it will be reused in the future by the Look up events: Search features.

### Tests
- Not yet, tests coming in future PR


peterinnesmsft:
Just as a general thing, we should pull in the common StyleCop/FxCop stuff that is being used elsewhere and apply it here if it's not being used already. This may just be an off-shoot of branching from the runtime code presented as is, but would be good to make sure we're applying the same overall standards as in our other repos.

PR #314, opened 2020-09-17
[CalendarV2] look up event - search

### LINKED ISSUE #304: Look up event(s): Search
## User Story
**As an** employee of an enterprise 
**I want** to ask VA to find a meeting for me, I will call out details that are top of mind such as start time and attendee (*could include title meeting if it’s easy to remember). 
**so that** I find the meeting I want to know more about and potentially take action

## Conversation Design
[Design](https://www.figma.com/file/xsSfVmzSe5CyFlcLovplS9/Calendar-Skill?node-id=26%3A1)
[Transcript](https://microsoft.sharepoint-df.com/:x:/t/BotFrameworkSolutions/Eer8FvZT2DhPmYMBfRVa7X4Bf9R0w6Bm9JJn5y8sd50g4A?e=iiHTaQ)
[Card design and considerations](https://www.figma.com/file/xsSfVmzSe5CyFlcLovplS9/Calendar-Skill?node-id=180%3A4)

## Language Understanding
| Intent | Sample Utterances | Prebuilt Entities | Custom Entities | Patterns |
| ------ | ----------------------| ----------------- | ----------------- | --------- |
GetEventTitle | | | | |
GetEventAttendees | | | | |
GetEventTime | | | | |
GetEventLocation | | | | |

## Dialogs
GetEventTitleDialog
- Collect parameters from LUIS entities. If not provided, prompt for which meeting.
- Call the SearchEventsDialog with parameters as options
- Display relevant LG response from Excel transcript 

GetEventAttendeesDialog
- Collect parameters from LUIS entities. If not provided, prompt for which meeting.
- Call the SearchEventsDialog with parameters as options
- Display relevant LG response from Excel transcript 

GetEventTimeDialog
- Collect parameters from LUIS entities. If not provided, prompt for which meeting.
- Call the SearchEventsDialog with parameters as options
- Display relevant LG response from Excel transcript 

GetEventLocationDialog
- Collect parameters from LUIS entities. If not provided, prompt for which meeting.
- Call the SearchEventsDialog with parameters as options
- Display relevant LG response from Excel transcript 

SearchEventsDialog
* Collect parameters from dialog options (title, datetime, and/or attendee).
* Call MS Graph – Search Events action with parameters
* Return results of API call in EndDialog action

## Custom Actions
| Name | Input | Output | Notes |
| ------ | ------ | -------- | ------- |
| MS Graph - Search Events | Start <br> End <br> Time zone <br> Title <br> Description <br> Attendees <br> **Ordinal (next)** | Collection of event objects | |

## State Properties
| Scope | Property | Type | Notes |
| ------- | --------- | ----- | ------- |
| user | token | TokenResponse object from OAuth Prompt | |
| user | settings.mailboxSettings.timeZone | Time zone string (i.e. Pacific Standard Time) from user's MS Graph MailBotSettings (see Get User Settings feature) | |
| conversation | events | Collection of event objects | |

## Channels
N/A

## Acceptance Criteria
- [ ] Flow should work in both Direct mode (connecting to the skill bot directly
~- [ ] Flow should work in Skill mode (connecting through a VA).~ This is on hold until a central VA can be created and deployed
- [ ] Should be deployed in a testing environment for review.
- [ ] Flow should work and render correctly in Teams on Desktop
- [ ] Flow should be tested in Teams for Mac and Teams for Mobile and any compatibility issues should be documented in a comment on this issue
- [ ] Flow should work and render correctly in Web Chat.
- [ ] Flow should work in Direct Line Speech and should follow the design~~ for speech scenarios.
- [ ] Each response (excluding card responses) should have at least 3 variant LG responses.
- [ ] Each response and card should include speak property provided by design team.
~- [ ] Help and Cancel interruptions should be implemented and customized for this dialog where relevant~ This requires greater all up design input
- [ ] Should include supporting unit tests for both custom actions and declarative dialogs.
- [ ] Language model should be tested against other intents to ensure that there is adequate training data to prevent conflicts
- [ ] Supported languages: English

xhr0804:
### Purpose
close #304 

Task List:
- [p0] I can search the meeting details based on 
- [x] start time
- [x] title
- [x] attendees 
- [p0] Search some specific infos includes 
- [x] EventTitle
- [x] EventAttendees
- [x] EventTime
- error path
- [x] [p0]No search condition
- [x] [p0]No available meetings based on condition (need designer review)
- [x] [p0]Still no meetings find after retry, show Day Overview
- [x] [p1] Multiple meeting item pick up, leverage the ListDialog


skills/declarative/calendarSkill/dialogs/SearchEventsDialog/SearchEventsDialog.dialog
```diff
@@ -0,0 +1,341 @@
...
+      "actions": [
+        {
+          "$kind": "Microsoft.TraceActivity",
+          "$designer": {
+            "id": "SapBYc"
+          },
+          "name": "SearchEventIn"
+        },
+        {
+          "$kind": "Microsoft.EmitEvent",
```
lauren-mills, line 25: I don't think we should introduce a new pattern here with the event. We can achieve the same thing by adding in the resolveDateTimeDialog to the intent triggers on calendarSkillDialog, and passing in the title and personNameArray, and datetime from the resolve dialog into SearchEvents as dialog options. Then just add the resolveDateTime and setProperties steps into the prompt in searchEvents.

skills/declarative/calendarSkill/dialogs/SearchEventsDialog/SearchEventsDialog.dialog
```diff
@@ -0,0 +1,341 @@
...
+              "validations": [
+                "@datetime!=null || @title!=null || @personName!=null || #no != null"
+              ]
+            },
+            {
+              "$kind": "Microsoft.IfCondition",
+              "$designer": {
+                "id": "F1YODj"
+              },
+              "condition": "#No != null",
```
lauren-mills, line 81: We should add a separate trigger for "no" here instead of doing it in line to make it more dynamic. You can set the allowInterruptions flag on the prompt to "=turn.recognized.score > 0.8" to keep it from over firing.

skills/declarative/calendarSkill/dialogs/SearchEventsDialog/SearchEventsDialog.dialog
```diff
@@ -0,0 +1,341 @@
...
+              "value": "= null"
+            }
+          ]
+        },
+        {
+          "$kind": "Microsoft.IfCondition",
+          "$designer": {
+            "id": "wRHP3J"
+          },
+          "condition": "dialog.events== null || count(dialog.events) == 0",
```
lauren-mills, line 142: I think this should be combined with the second if/else block below to condense it and make it a little easier to follow.

lauren-mills (changes requested):

lauren-mills:
I noticed that this doesn't enforce selecting one result, which will be something we need for the update and cancel dialogs. If the selection property is passed into the dialog options, the searchEvents dialog should ask the user to choose only one result and return only one. That was the purpose of separating out SearchEvents from the GetEventsByDate dialog so that logic could be added.

skills/declarative/calendarSkill/calendarskill.dialog
```diff
@@ -164,9 +171,185 @@
...
+            "from": "GetEventLocation"
+          }
+        }
+      ]
     }
   ],
   "$schema": "https://raw.githubusercontent.com/microsoft/BotFramework-Composer/stable/Composer/packages/server/schemas/sdk.schema",
   "generator": "calendarskill.lg",
-  "recognizer": "calendarskill.lu"
+  "recognizer": "calendarskill.lu.qna",
```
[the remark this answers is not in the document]
xhr0804: I think this is generated automatically when using the latest composer version v1.1.1. The latest composer may generate this pattern for all the dialog?

skills/declarative/calendarSkill/dialogs/AuthenticationDialog/AuthenticationDialog.dialog
```diff
@@ -48,5 +48,7 @@
       ]
     }
   ],
-  "generator": "AuthenticationDialog.lg"
+  "generator": "AuthenticationDialog.lg",
+  "id": "AuthenticationDialog",
+  "recognizer": "AuthenticationDialog.lu.qna"
```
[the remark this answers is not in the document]
xhr0804: Same with above, this is generated automatically when using composer v1.1.1, should we align to it?

skills/declarative/calendarSkill/calendarskill.dialog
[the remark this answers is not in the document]
lauren-mills: What branch of composer are you working from? I'm using master for consistency, so we may need to align here.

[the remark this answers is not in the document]
xhr0804: @lauren-mills I'm using the composer released client v1.1.1 https://github.com/microsoft/BotFramework-Composer/releases. And Yue @KayMKM is using the main branch and looks like the composer team doesn't update the master branch anymore.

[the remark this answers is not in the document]
lauren-mills: Correction, I am also using the main branch, I mistyped.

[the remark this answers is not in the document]
lauren-mills: Let's all switch to the 1.1.1 client for consistency

lauren-mills:
I took another look at it and talked to design, and I think from a user experience perspective the logic will be easier to follow if the SearchEventsDialog doesn't handle the final response with the IF/ELSE LG template based on intent, but instead it should return 1 or more events in the EndDialog step and let have the calling trigger off of calendarSkill handle the response. This will be more helpful to the users when they are trying to make changes to how the skill responds to a given intent.



This also helps keep the SearchEventsDialog clean, because it will accept the different entities as parameters, perform the search, then return the result without being concerned with how the result is rendered. It makes it a one purpose component which is easier to understand for people who might use the sklil as a sample later.

skills/declarative/calendarSkill/dialogs/SearchEventsDialog/SearchEventsDialog.dialog
```diff
@@ -0,0 +1,373 @@
...
+          "actions": [
+            {
+              "$kind": "Microsoft.TraceActivity",
+              "$designer": {
+                "id": "1eO2VL"
+              },
+              "name": "CollectOption"
+            },
+            {
+              "$kind": "Microsoft.DeleteProperties",
```
lauren-mills, line 52: You should be able to remove this step by setting "Always prompt" to true on the text input

skills/declarative/calendarSkill/dialogs/SearchEventsDialog/SearchEventsDialog.dialog
```diff
@@ -0,0 +1,373 @@
...
+          "attendeesProperty": "=coalesce(dialog.personNameArray, null)",
+          "startProperty": "=coalesce(dialog.start, formatDateTime(utcNow(), 'yyyy-MM-dd'))",
+          "endProperty": "=coalesce(dialog.end, null)",
+          "timeZoneProperty": "=user.workingHours.timeZone.name",
+          "isFutureProperty": "=coalesce(dialog.futureEventsOnly, false)",
+          "isFutureEventsOnlyProperty": "=dialog.futureEventsOnly",
+          "dateTimeTypeProperty": "= coalesce(dialog.type, \"date\")"
+        },
+        {
+          "$kind": "Microsoft.SetProperties",
```
lauren-mills, line 104: Why did you need to set this to null here?

skills/declarative/calendarSkill/dialogs/SearchEventsDialog/SearchEventsDialog.dialog
```diff
@@ -0,0 +1,373 @@
...
+                    {
+                      "property": "dialog.events[dialog.foreach.index].Location",
+                      "value": "= dialog.events[dialog.foreach.index].Location.DisplayName"
+                    }
+                  ]
+                }
+              ]
+            },
+            {
+              "$kind": "Microsoft.IfCondition",
```
lauren-mills, line 214: The dialog should be able to handle returning either a list of events or a single event depending on how its configured in dialog options. So if you have a flag called "doSelection" for example you could show the choice prompt if its true, or return the full results list if its false. That way we could reuse this dialog for the Day Overview as well

skills/declarative/calendarSkill/calendarskill.dialog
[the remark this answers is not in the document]
xhr0804: Sure, let's do that

skills/declarative/calendarSkill/dialogs/SearchEventsDialog/SearchEventsDialog.dialog
xhr0804: Sure, removed this pattern and simplified the logic

skills/declarative/calendarSkill/dialogs/SearchEventsDialog/SearchEventsDialog.dialog
xhr0804: Fixed

skills/declarative/calendarSkill/dialogs/SearchEventsDialog/SearchEventsDialog.dialog
xhr0804: Removed for the latest flow. It was because when I repeated the dialog, the luis result would influence the flow.

skills/declarative/calendarSkill/dialogs/SearchEventsDialog/SearchEventsDialog.dialog
xhr0804: Sure, modified the flow to match this pattern.

skills/declarative/calendarSkill/calendarskill.dialog
```diff
@@ -180,24 +180,24 @@
...
           "$kind": "Microsoft.BeginDialog",
           "$designer": {
-            "id": "9n2omv"
+            "id": "ESk1DO"
           },
           "activityProcessed": true,
-          "dialog": "AuthenticationDialog"
+          "dialog": "ResolveDateTimeDialog",
+          "options": {
+            "query": "=turn.activity.text"
```
lauren-mills, line 198: The setProperty for the query property actually has to stay for scenarios when the user isn't authenticated yet when they call this dialog (through a VA for instance). If the authentication dialog requires a magic code, the turn.activity.text won't have the user's original utterance anymore

skills/declarative/calendarSkill/dialogs/SearchEventsDialog/SearchEventsDialog.dialog
```diff
@@ -0,0 +1,267 @@
...
+              "itemsProperty": "dialog.events",
+              "actions": [
+                {
+                  "$kind": "Microsoft.SetProperties",
+                  "$designer": {
+                    "id": "M5nVxN"
+                  },
+                  "assignments": [
+                    {
+                      "property": "dialog.events[dialog.foreach.index].Location",
```
lauren-mills, line 167: I think we can make this change in the getEvents or the sortEvents action to remove the need for this. I'll make this change and make a PR

lauren-mills: Actually, since I don't have your getEvents action changes, I'll let you make the change. Here is the method you can use to take the raw event payload and convert it into the schema I used for the card:


        private List<object> ParseEvents(DateTime currentDateTime, List<Event> events)
        {
            var eventsList = new List<object>();
            var i = 0;

            foreach (var ev in events)
            {
                var start = DateTime.Parse(ev.Start.DateTime);
                var end = DateTime.Parse(ev.End.DateTime);
                var duration = end.Subtract(start);
                var isCurrentEvent = false;

                if(start <= currentDateTime && currentDateTime <= end
                    || start.AddMinutes(-30) <= currentDateTime && currentDateTime <= start)
                {
                    // If event is currently ongoing, or will start in the next 30 minutes
                    isCurrentEvent = true;
                }

                eventsList.Add(new
                {
                    Id = i++,
                    ev.Subject,
                    ev.Start,
                    ev.End,
                    ev.Attendees,
                    ev.IsOnlineMeeting,
                    ev.OnlineMeeting,
                    Description = ev.BodyPreview,
                    Location = !string.IsNullOrEmpty(ev.Location.DisplayName) ? ev.Location.DisplayName : string.Empty,
                    DurationDays = duration.Days,
                    DurationHours = duration.Hours,
                    DurationMinutes = duration.Minutes,
                    isRecurring = ev.Type == EventType.Occurrence || ev.Type == EventType.SeriesMaster ? true : false,
                    isCurrentEvent,
                    ev.IsOrganizer,
                    ev.ResponseStatus.Response,
                    ev.Organizer
                });
            }

            return eventsList;
        }

xhr0804: GetEvent is called before SortEvent, if I change the output of GetEvent, something will be wrong with SortEvent and maybe other parts. So I changed the SortEvent to handle two scenarios, one returns events grouped by date and the other returns sorted events directly. Both convert the event to the schema used in card.

skills/declarative/calendarSkill/calendarskill.dialog
xhr0804: OK, fixed

xhr0804:
@lauren-mills Here are some scenerios that the ListDialog doesn't support.
1. After user choose one meeting, bot says something like "- Sure. Let’s continue with the [#] event." We may need to add a template for this response.
2. After user get to the end of the list, a template are also needed for response like "- That's it, you're at the end of the list. Do you want to know what’s on today or create a new meeting?"
3. According to the design, we need to limit the total numbers to 25-items or 5-pages. We should support "Maximum" feature.

lauren-mills:
@xhr0804 - For the first one, I think the calling dialog should handle it. For the other two, I agree and I'll work on adding them in!

xhr0804:
@lauren-mills But for the first one, calling dialog can not get the **exact event number ([#] event)** which would be used in the response like "- Sure. Let’s continue with the **[#] event**.".

lauren-mills (approved):

PR #318, opened 2020-09-24
[Calendar v2] choose contact dialog 

KayMKM:
### Purpose
1. add a choose contact dialog whose input is a list of contact candidates and output is a attendee item.
2. discuss about how to reuse this dialog in choose time suggestion and choose meeting

### Changes
image
image
image
image
image

### Tests

### Feature Plan

#### General


Quirinevwm:
Nice, the list works smoothly! I see some minor things I would like to update from design and transcript perspective:
- include a comma between [Name], [email address]
- confirmation transcript is missing after providing options, I've included in the transcript the following: 
image
- plus we have new transcript for ending of a list (we discussed yesterday)
- open question I have will we include the action item on the card for "try again" - it's there in the design but not in the test

lauren-mills:
I think we can refactor this this put the logic for 'next', 'previous', etc all in the interruption triggers themselves rather than just setting the interruptReason. That would make each dialog easier to follow.

Also, I think we can make this a generic ChooseOption dialog, by passing in the parameters for the adaptive card (title, header image url, button options. etc) in as dialogOptions. Then we can just call this same dialog from each of the of the other dialogs when we need to handle lists and we'll only have one LUIS app deployed to support the next, previous, showMore intents.

KayMKM:
@lauren-mills I have refactored this dialog to ChooseOption. To use this dialog uesr could pass card header, icon, the name of item card template, action template, prompt, reprompt etc. And I use template() to generate the lg with template name, but seems it can only call the template in the same scope. So I put the prompt/reprompt lg in ChooseOption.lg as workaround.

@Quirinevwm I fixed these issues and updated the demo transcript.

KayMKM:
> I use template() to generate the lg with template name, but seems it can only call the template in the same scope. So I put the prompt/reprompt lg in ChooseOption.lg as workaround.
> 

Imported GetContactDialog.lg in ChooseOption.lg to fix this issue

PR #412, opened 2020-11-19
Introduce a feed controlling which components to display in Composer.

### LINKED ISSUE #411: Introduce a feed controlling which components to display in Composer.
**Is your feature request related to a problem? Please describe.**

Composer's upcoming Package Manager feature displays a list of first party components based on a JSON feed pulled from the web. I would like to host this feed in this repo, since many of the 1p components will also be here.

**Describe the solution you'd like**

I would like to introduce `feeds/components.json` file.  In the future, this file would be updated by submitting pull requests to the repo.

benbrown:
Closes #411 

### Purpose
This introduces `feeds/components.json` which contains a list of the 4 current "core" components at their preview locations.

This feed will be consumed by Composer 1.3's Package Manager extension. Future versions of Composer may not require this file.


lauren-mills (approved):

benbrown:
The expected form of this feed can be found here.

https://github.com/microsoft/BotFramework-Composer/blob/main/extensions/packageManager/src/client/libraryList.tsx#L27

feeds/components.json
```diff
@@ -0,0 +1,46 @@
+[
+  {
+    "name": "Preview.ConversationalCore.GeneralIntent",
```
peterinnesmsft, line 3: A lot of this information seems duplicative of metadata that should naturally already be within a .nuspec or package.json file. Would it not make more sense to largely have this be pointers to some backing package from whichever source, and then normalize the package metadata into a standard representation for the purposes of rendering in Composer, etc.?

feeds/components.json
```diff
@@ -0,0 +1,46 @@
...
+    "version": "0.0.1-preview1",
```
peterinnesmsft, line 4: version feels like it ought to be an array or capable of representing version ranges. Or indeed, do we even need this, and instead just pull all available versions available from the backing package source?

peterinnesmsft (commented):
Seems like we could minimize duplication of information that ought to be in the backing package already. Would also be good to see if we can streamline versions a little bit.

feeds/components.json
benbrown: This is should be understood as the latest version. Users can install older versions by hand.

Eventually this data will come from nuget or npm, where there will be a single version represented in the feed.

feeds/components.json
benbrown: Ultimately we can get a lot of this information from an npm search, which does in turn get the data from the packages.  We will do this in an upcoming version of Composer once components are regularly being published under out controlled namespace and a tagging strategy has been determined.

For example:
https://registry.npmjs.org/-/v1/search?text=botbuilder&keywords:botframework&size=100&from=0&quality=0.65&popularity=0.98&maintenance=0.5

Meanwhile, having the information here takes the burden off the consumer to dereference each package and extract the information, something that would be very time consuming to do at runtime.

feeds/components.json
benbrown: This _can be_ optional in the feed fwiw, in which case it pulls the latest version.

benbrown: however listing the version makes the table look better in Composer ;)

feeds/components.json
peterinnesmsft: I don't think it can be *that* time consuming surely? Given we have identifiers for each package, I can't imagine that hydrating the metadata per package can be more than ~250-500ms, which if you then throw parallelization and caching into the mix would eliminate the need for repeated retrievals. That ought to be a very negligible cost, particularly if it means I only ultimately have to specify my metadata in one location.

feeds/components.json
peterinnesmsft: I think I'm even more confused then - given my other comment, as I've understood the response data for retrieving package metadata from both npm and NuGet, I believe they largely contain all of the available versions. At that point, would we not just display the latest version as available directly from the source?

peterinnesmsft (approved):

feeds/components.json
peterinnesmsft: Spoke offline - intention is to for 1.4 release to pull entire feed.

clearab (approved):

PR #419, opened 2020-11-30
[CalendarV2] fix bugs

### LINKED ISSUE #371: Test/Bug: Search for meeting at a certain time works well, maybe we can include if it's organized by yourself that we show - "you are the organizer"
Search for meeting at a certain time works well, maybe we can include if it's organized by yourself that we show - "you are the organizer"
image

### LINKED ISSUE #402: [Calendar Skill V2] suggested time could overly navigate
## Describe the bug

After clicking the > arrow, it could navigate to a bad page and meet errors:
image

### LINKED ISSUE #417: [Calendar Skill V2] internal exchange address is not properly displayed
## Describe the bug
If it is an exchange address, it is not properly displayed:
image

## To Reproduce
Create a contact that refers an internal user and then search

## Additional context
It seems it could be resolved like this https://social.msdn.microsoft.com/Forums/office/en-US/e78bd9ad-4ec3-4f2d-9939-747cafe77faa/how-to-resolve-the-email-address-in-this-form-oabcdouexchange-administrative-group

xhr0804:
### Purpose
close #417 
close #371
close #402 

### Changes

### Tests

### Feature Plan

#### General

### DIFF
--- a/skills/declarative/calendarSkill/dialogs/ListDialog/ListDialog.dialog
+++ b/skills/declarative/calendarSkill/dialogs/ListDialog/ListDialog.dialog
@@ -109,11 +109,17 @@
           ]
         },
         {
+          "$kind": "Microsoft.TraceActivity",
+          "$designer": {
+            "id": "tdNJ28"
+          }
+        },
+        {
           "$kind": "Microsoft.IfCondition",
           "$designer": {
             "id": "P1awzy"
           },
-          "condition": "(dialog.pageIndex * dialog.pageSize) - count(dialog.Items) <= dialog.pageSize",
+          "condition": "dialog.pageIndex * dialog.pageSize <  count(dialog.Items)",
           "actions": [
             {
               "$kind": "Microsoft.RepeatDialog",
@@ -132,6 +138,14 @@
             }
           ],
           "elseActions": [
+            {
+              "$kind": "Microsoft.SetProperty",
+              "$designer": {
+                "id": "lepsYO"
+              },
+              "property": "dialog.pageIndex",
+              "value": "= dialog.pageIndex - 1"
+            },
             {
               "$kind": "Microsoft.SendActivity",
               "$designer": {
@@ -191,6 +205,14 @@
             }
           ],
           "elseActions": [
+            {
+              "$kind": "Microsoft.SetProperty",
+              "$designer": {
+                "id": "MgUdSM"
+              },
+              "property": "dialog.pageIndex",
+              "value": "= 0"
+            },
             {
               "$kind": "Microsoft.SendActivity",
               "$designer": {
--- a/skills/declarative/calendarSkill/language-generation/en-us/event.en-us.lg
+++ b/skills/declarative/calendarSkill/language-generation/en-us/event.en-us.lg
@@ -294,7 +294,7 @@
         ]
     },
     ```
-- ELSEIF: ${ event.Attendees != null && count(event.Attendees) == 1 && event.Response != 'organizer'}
+- ELSEIF: ${ event.Attendees != null && count(event.Attendees) == 1}
     -  ```
     {
         "type": "ColumnSet",
--- a/skills/declarative/calendarSkill/language-generation/en-us/event.en-us.lg
+++ b/skills/declarative/calendarSkill/language-generation/en-us/event.en-us.lg
@@ -259,7 +259,9 @@
     -
 
 # EventAttendeeContainer(event)
-- IF: ${ event.attendees != null && count(event.attendees) > 1 }
+- IF: ${ event.Response == 'organizer' }  
+    -
+- ELSEIF: ${ event.attendees != null && count(event.attendees) > 1 }
     - ```
     {
         "type": "ColumnSet",
--- a/skills/declarative/calendarSkill/language-generation/en-us/event.en-us.lg
+++ b/skills/declarative/calendarSkill/language-generation/en-us/event.en-us.lg
@@ -259,9 +259,7 @@
     -
 
 # EventAttendeeContainer(event)
-- IF: ${ event.Response == 'organizer' }  
-    -
-- ELSEIF: ${ event.attendees != null && count(event.attendees) > 1 }
+- IF: ${ event.attendees != null && count(event.attendees) > 1 }
     - ```
     {
         "type": "ColumnSet",
@@ -296,7 +294,7 @@
         ]
     },
     ```
-- ELSEIF: ${ event.Attendees != null && count(event.Attendees) == 1}
+- ELSEIF: ${ event.Attendees != null && count(event.Attendees) == 1 && event.Response != 'organizer'}
     -  ```
     {
         "type": "ColumnSet",
--- a/skills/declarative/calendarSkill/runtime/customaction/Actions/MSGraph/GetContacts.cs
+++ b/skills/declarative/calendarSkill/runtime/customaction/Actions/MSGraph/GetContacts.cs
@@ -95,6 +95,7 @@ namespace Microsoft.BotFramework.Composer.CustomAction.Actions.MSGraph
 
             if (people?.Count > 0)
             {
+                var existingResult = contactsResult.SelectMany(c => c.EmailAddresses).ToHashSet();
                 foreach (var person in people)
                 {
                     var emailAddresses = new List<string>();
@@ -102,7 +103,7 @@ namespace Microsoft.BotFramework.Composer.CustomAction.Actions.MSGraph
                     foreach (var email in person.ScoredEmailAddresses)
                     {
                         // If the email address isn't already included in the contacts list, add it
-                        if (!contactsResult.SelectMany(c => c.EmailAddresses).Contains(email.Address) && IsEmail(email.Address))
+                        if (!existingResult.Contains(email.Address) && IsEmail(email.Address))
                         {
                             emailAddresses.Add(email.Address);
                         }
--- a/skills/declarative/calendarSkill/runtime/customaction/Actions/MSGraph/GetContacts.cs
+++ b/skills/declarative/calendarSkill/runtime/customaction/Actions/MSGraph/GetContacts.cs
@@ -13,6 +13,7 @@ using Microsoft.BotFramework.Composer.CustomAction.Models;
 using Microsoft.Graph;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
+using AdaptiveExpressions;
 
 namespace Microsoft.BotFramework.Composer.CustomAction.Actions.MSGraph
 {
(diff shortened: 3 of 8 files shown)


skills/declarative/calendarSkill/runtime/customaction/Actions/MSGraph/GetContacts.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Net.Http;
+using System.Text.RegularExpressions;
```
thomas-chung, line 5: nit: if we order the using's alphabetically we can eliminate a lot of the compiler warnings + make the code easier to read.

skills/declarative/calendarSkill/runtime/customaction/Actions/MSGraph/GetContacts.cs
```diff
@@ -63,23 +65,23 @@ public override async Task<DialogTurnResult> BeginDialogAsync(DialogContext dc,
             {
                 foreach (var contact in contacts)
                 {
-                    var emailAddresses = new List<string>();
-
-                    foreach (var email in contact.EmailAddresses)
+                    //var emailAddresses = new List<string>();
```
thomas-chung, line 68: Dead code, probably can just go ahead and remove.

skills/declarative/calendarSkill/runtime/customaction/Actions/MSGraph/GetContacts.cs
```diff
@@ -134,5 +136,11 @@ public override async Task<DialogTurnResult> BeginDialogAsync(DialogContext dc,
             // return the actionResult as the result of this operation
             return await dc.EndDialogAsync(result: results, cancellationToken: cancellationToken);
         }
+
+        private bool IsEmail(string emailString)
+        {
+            bool isEmail = !string.IsNullOrEmpty(emailString) && Regex.IsMatch(emailString, @"\A(?:[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*@(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?)\Z", RegexOptions.IgnoreCase);
```
thomas-chung, line 142: For perf gain you might want to also compile the regex as well.

skills/declarative/calendarSkill/runtime/customaction/Actions/MSGraph/GetContacts.cs
```diff
@@ -63,23 +65,23 @@ public override async Task<DialogTurnResult> BeginDialogAsync(DialogContext dc,
             {
                 foreach (var contact in contacts)
                 {
-                    var emailAddresses = new List<string>();
-
-                    foreach (var email in contact.EmailAddresses)
+                    //var emailAddresses = new List<string>();
+                    var emailAddresses = contact.EmailAddresses.Where(e => !string.IsNullOrEmpty(e.Address) && IsEmail(e.Address)).Select(e => e.Address).ToList();
```
thomas-chung, line 69: The IsNullOrEmpty is sort of redundant here because IsEmail already checks it.

skills/declarative/calendarSkill/runtime/customaction/Actions/MSGraph/GetContacts.cs
```diff
@@ -102,7 +104,7 @@ public override async Task<DialogTurnResult> BeginDialogAsync(DialogContext dc,
                     foreach (var email in person.ScoredEmailAddresses)
                     {
                         // If the email address isn't already included in the contacts list, add it
-                        if (!contactsResult.SelectMany(c => c.EmailAddresses).Contains(email.Address))
+                        if (!contactsResult.SelectMany(c => c.EmailAddresses).Contains(email.Address) && IsEmail(email.Address))
```
thomas-chung, line 107: This way of writing it is not optimal for large sets because we are asking the runtime to basically regenerate a stable list for every scored email address we have in the list which is an O(M*N) operation followed by another O(N) operation (the Contains()). 

I think we can make this code run faster, we can have a HashSet<T> of email addresses with the Contains() which turns it into an O(N) operation, then an O(1).

skills/declarative/calendarSkill/language-generation/en-us/event.en-us.lg
```diff
@@ -259,7 +259,9 @@
     -
 
 # EventAttendeeContainer(event)
-- IF: ${ event.attendees != null && count(event.attendees) > 1 }
+- IF: ${ event.Response == 'organizer' }
```
thomas-chung, line 262: New person question - can we just eliminate this if and instead have 

"IF ${event.Response != 'organizer' && event.attendees != null && count(event.attendees) > 1}"

Maybe it's clearer that way

skills/declarative/calendarSkill/language-generation/en-us/event.en-us.lg
```diff
@@ -259,7 +259,9 @@
     -
 
 # EventAttendeeContainer(event)
-- IF: ${ event.attendees != null && count(event.attendees) > 1 }
+- IF: ${ event.Response == 'organizer' }  
+    -
+- ELSEIF: ${ event.attendees != null && count(event.attendees) > 1 }
```
thomas-chung, line 264: Dumb question - is event.attendees > 1 because the organizer is always person 0?

skills/declarative/calendarSkill/dialogs/ListDialog/ListDialog.dialog
```diff
@@ -108,12 +108,18 @@
...
+            "id": "tdNJ28"
+          }
+        },
         {
           "$kind": "Microsoft.IfCondition",
           "$designer": {
             "id": "P1awzy"
           },
-          "condition": "(dialog.pageIndex * dialog.pageSize) - count(dialog.Items) <= dialog.pageSize",
+          "condition": "dialog.pageIndex * dialog.pageSize <  count(dialog.Items)",
```
thomas-chung, line 122: I wonder if this equation is "too much math gymnasium" for non-technical customers to understand and whether or not we want to simplify this further. 

Basically what we are saying is "given the result set, what is the pages required to display all the information." then there is the question of "given where we are in the page index, are we at the end yet?" If we model it this way would it be easier to understand? And should we do it this way too?

skills/declarative/calendarSkill/dialogs/ListDialog/ListDialog.dialog
```diff
@@ -191,6 +205,14 @@
             }
           ],
           "elseActions": [
+            {
+              "$kind": "Microsoft.SetProperty",
```
thomas-chung, line 209: I wonder if we should challenge this way to doing the flow design.

This makes complete sense to a programmer like us. But if we are shipping this to customers who are just dragging and dropping, then would they be overwhelmed and spend way too much time wondering why this is there in the first place?

Basically the way I think about this is - the dialog has a state - the current index, the number of pages required to display all the results. The various events/intents (still trying to use the right terms here) just trying to advance this count +1 and -1 until it hits either the beginning or end. Can we model it in such a way that the dialog's flow can demonstrate this?

thomas-chung: By the way, pagination is such an interesting reusable thing (I think). We may want to drill down on it some more.

skills/declarative/calendarSkill/runtime/customaction/Actions/MSGraph/GetContacts.cs
lauren-mills: Yes, we need to add in the stylecop ruleset to these projects. I'll add that as a separate work item

skills/declarative/calendarSkill/dialogs/ListDialog/ListDialog.dialog
lauren-mills: @thomas-chung - why don't you take a look at the ListDialog separately and try to model it this way? It'd be a good exercise since that is one of the components we're hoping to ship as a separate reusable asset

skills/declarative/calendarSkill/language-generation/en-us/event.en-us.lg
lauren-mills: I don't think this logic is quite right. If the user is the organizer, this will prevent it from showing any of the other attendees, when really we only want to hide the attendees if the user is the organizer and there are no other people coming to the meeting.

So the condition that checks for count(event.Attendees) == 1 should also include "&& event.Response != 'organizer"'.

lauren-mills (changes requested):

skills/declarative/calendarSkill/runtime/customaction/Actions/MSGraph/GetContacts.cs
thomas-chung: Actually our build already generates warnings for these types of issue. At least that's what I see when we are compiling main.

xhr0804: Fixed this line. I think there must be lots of style warning in this project. We can fix all at once later separately.

skills/declarative/calendarSkill/runtime/customaction/Actions/MSGraph/GetContacts.cs
xhr0804: Added a RegexOptions.Compiled. Does that work?

skills/declarative/calendarSkill/language-generation/en-us/event.en-us.lg
xhr0804: Actually, the previous logic is enough for most cases except a special one, that is the event.Attendees sometimes include the orgnizer him/herself and sometimes not include. So even if count(event.Attendees) ==1, we couldn't know whether the attendee is the orgnizer him/herself.

xhr0804:
@lauren-mills Added a logic to filter the event.Attendees with the current user's Email. This will ensure the attendee list just contains other people and be empty if no other person.

skills/declarative/calendarSkill/runtime/customaction/Actions/MSGraph/GetContacts.cs
xhr0804: good point, updated the code.

lauren-mills (approved):

skills/declarative/calendarSkill/runtime/customaction/Actions/MSGraph/GetContacts.cs
```diff
@@ -134,5 +135,12 @@ public override async Task<DialogTurnResult> BeginDialogAsync(DialogContext dc,
             // return the actionResult as the result of this operation
             return await dc.EndDialogAsync(result: results, cancellationToken: cancellationToken);
         }
+
+        private bool IsEmail(string emailString)
+        {
+
```
thomas-chung, line 141: nit: extra line

thomas-chung (approved):
> Fixed this line. I think there must be lots of style warning in this project. We can fix all at once later separately. 

Yeah actually there are. I agree we can fix those later.

thomas-chung (approved):

skills/declarative/calendarSkill/runtime/customaction/Actions/MSGraph/GetContacts.cs
xhr0804: Oops..fixed

PR #450, opened 2020-12-09
Refactor MS Graph code in Calendar skill v2

thomas-chung:
### Purpose
Refactor MS Graph related code to use common base class and eliminate component registration requirement via attribute on the custom action

### Changes
- All the MS Graph calls are now internal to the base class
- Consumer of the base class would override a single method to return the result (logging, error handling are also in base class now)
- New ComponentRegistration attribute to automate registration of the custom action in the runtime

### Tests
It's a refactor

### Feature Plan
No

#### General

### DIFF
--- a/skills/declarative/calendarSkill/runtime/customaction/Actions/MSGraph/CheckAvailability.cs
+++ b/skills/declarative/calendarSkill/runtime/customaction/Actions/MSGraph/CheckAvailability.cs
@@ -1,17 +1,16 @@
-﻿using AdaptiveExpressions.Properties;
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using AdaptiveExpressions.Properties;
 using Microsoft.Bot.Builder.Dialogs;
 using Microsoft.Bot.Builder.TraceExtensions;
-using Microsoft.Recognizers.Text.DataTypes.TimexExpression;
-using Microsoft.Recognizers.Text.DateTime;
-using Microsoft.Graph;
+using Microsoft.BotFramework.Composer.CustomAction.Models;
 using Newtonsoft.Json;
 using System;
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
-using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.BotFramework.Composer.CustomAction.Models;
 
 namespace Microsoft.BotFramework.Composer.CustomAction.Actions.MSGraph
 {
@@ -20,9 +19,9 @@ namespace Microsoft.BotFramework.Composer.CustomAction.Actions.MSGraph
     /// This has proven to be more consistent that LUIS datetime recognition by allowing the use of the 
     /// user's current timezone time as a relative datetime rather than a mix of absolute times and UTC times.
     /// </summary>
+    [ComponentRegistration(CheckAvailability.DeclarativeType)]
     public class CheckAvailability : Dialog
     {
-        [JsonProperty("$kind")]
         public const string DeclarativeType = "Microsoft.Graph.Calendar.CheckAvailability";
 
         [JsonConstructor]
--- a/skills/declarative/calendarSkill/runtime/customaction/Actions/MSGraph/CheckAvailability.cs
+++ b/skills/declarative/calendarSkill/runtime/customaction/Actions/MSGraph/CheckAvailability.cs
@@ -19,9 +19,9 @@ namespace Microsoft.BotFramework.Composer.CustomAction.Actions.MSGraph
     /// This has proven to be more consistent that LUIS datetime recognition by allowing the use of the 
     /// user's current timezone time as a relative datetime rather than a mix of absolute times and UTC times.
     /// </summary>
-    [MsGraphCustomActionRegistration(CheckAvailability.DeclarativeType)]
     public class CheckAvailability : Dialog
     {
+        [JsonProperty("$kind")]
         public const string DeclarativeType = "Microsoft.Graph.Calendar.CheckAvailability";
 
         [JsonConstructor]
--- a/skills/declarative/calendarSkill/runtime/customaction/Actions/MSGraph/CheckAvailability.cs
+++ b/skills/declarative/calendarSkill/runtime/customaction/Actions/MSGraph/CheckAvailability.cs
@@ -19,7 +19,7 @@ namespace Microsoft.BotFramework.Composer.CustomAction.Actions.MSGraph
     /// This has proven to be more consistent that LUIS datetime recognition by allowing the use of the 
     /// user's current timezone time as a relative datetime rather than a mix of absolute times and UTC times.
     /// </summary>
-    [ComponentRegistration(CheckAvailability.DeclarativeType)]
+    [MsGraphCustomActionRegistration(CheckAvailability.DeclarativeType)]
     public class CheckAvailability : Dialog
     {
         public const string DeclarativeType = "Microsoft.Graph.Calendar.CheckAvailability";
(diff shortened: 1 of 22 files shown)


ryanlengel (commented):
Not required for now but we should consider what the ultimate project name of this should be and how many packages it is comprised of. Is it a combination of MS Graph & generiic custom actions?

skills/declarative/calendarSkill/runtime/customaction/Actions/MSGraph/CheckAvailability.cs
```diff
@@ -20,9 +19,9 @@ namespace Microsoft.BotFramework.Composer.CustomAction.Actions.MSGraph
     /// This has proven to be more consistent that LUIS datetime recognition by allowing the use of the 
     /// user's current timezone time as a relative datetime rather than a mix of absolute times and UTC times.
     /// </summary>
+    [MsGraphCustomActionRegistration(CheckAvailability.DeclarativeType)]
```
lauren-mills, line 22: This doesn't look like its using the same approach to setting the DeclarativeType as the other actions

skills/declarative/calendarSkill/runtime/customaction/MSGraphClient.cs
```diff
@@ -27,7 +33,9 @@ public static GraphServiceClient GetAuthenticatedClient(string accessToken, Http
 
         public static Exception HandleGraphAPIException(ServiceException ex)
         {
-            return new Exception($"Microsoft Graph API Exception: {ex.Message}");
+            // TODO: Why do we throw a very generic exception when there is a much better
```
lauren-mills, line 36: I believe this is a relic of the original C# implementation. We can take it out/modify this logic however we see fit. I think the original use of handling the ServieException differently was to be able to throw a specific message to the user like "I'm having trouble accessing Microsoft Graph" or something rather than "Sorry, something went wrong" but we haven't implemented that in this design

lauren-mills (changes requested):
This might be a good opportunity to add in the stylecop ruleset as well. I think we should only add it to the customActions project since everything else will change with the runtime work.

thomas-chung:
> Not required for now but we should consider what the ultimate project name of this should be and how many packages it is comprised of. Is it a combination of MS Graph & generiic custom actions?

@ryanlengel For now, I think we should just have one custom action package called "Microsoft.Bots.CustomActions.MsGraph" and then we can consider splitting them up at later time if necessary. This is for simplicity sake. Later point, if we wanted to split this up further we can append to the name like "Microsoft.Bots.CustomActions.MsGraph.Mails" for example. 

I think the naming model should follow that of other .NET Core components that we ship via nuget for consistency.

lauren-mills:
@thomas-chung, @ryanlengel 

I think what we may end up with for calendar is something like this package-wise:

- 1 package with just the MS Graph actions as close to the APIs as possible
- Calendar content package (dialog, LU, LG) plus **additional actions** needed to parse the results of the MS graph actions for the calendar specific stuff. These might include stuff like special sorting or conversion logic needed for the UI rendering etc. But would be decoupled from the actual graph API call.

What do you both think?

ryanlengel:
@lauren-mills that makes sense - additionally we should pull out the `GoToAction`, `RecognizeDateTime` & `SendActivityPlus` actions as they are appear generic and not calendar-specific. Is that correct?

lauren-mills:
Yes, GoToAction and SendActivity+ should ultimately be in the SDK but we needed our own copies until they're in and the DateTime action will go into a separate DateTimeResolution package that's standalone and has that action as well as a supporting dialog

thomas-chung:
> @thomas-chung, @ryanlengel
> 
> I think what we may end up with for calendar is something like this package-wise:
> 
> * 1 package with just the MS Graph actions as close to the APIs as possible
> * Calendar content package (dialog, LU, LG) plus **additional actions** needed to parse the results of the MS graph actions for the calendar specific stuff. These might include stuff like special sorting or conversion logic needed for the UI rendering etc. But would be decoupled from the actual graph API call.
> 
> What do you both think?

I think that's a good start. LU and LG can definitely be decoupled from the dialog. I think additional custom actions can stay within the main skills project since they are pretty much specific to the skill. This way we get a demonstration of custom actions internal and external (package), and then a variety of different assets that are also componentize.

thomas-chung:
> > @thomas-chung, @ryanlengel
> > I think what we may end up with for calendar is something like this package-wise:
> > 
> > * 1 package with just the MS Graph actions as close to the APIs as possible
> > * Calendar content package (dialog, LU, LG) plus **additional actions** needed to parse the results of the MS graph actions for the calendar specific stuff. These might include stuff like special sorting or conversion logic needed for the UI rendering etc. But would be decoupled from the actual graph API call.
> > 
> > What do you both think?
> 
> I think that's a good start. LU and LG can definitely be decoupled from the dialog. I think additional custom actions can stay within the main skills project since they are pretty much specific to the skill. This way we get a demonstration of custom actions internal and external (package), and then a variety of different assets that are also componentize.

The other thing too is now we are getting into the area of contracts. Because dialog makes reference to LU and LG packages, so we can in theory rev them separately. We should probably establish a "contract" to ensure updating the LU/LU packages don't end up breaking the bot.

skills/declarative/calendarSkill/runtime/customaction/Actions/MSGraph/CheckAvailability.cs
```diff
@@ -19,9 +19,9 @@ namespace Microsoft.BotFramework.Composer.CustomAction.Actions.MSGraph
     /// This has proven to be more consistent that LUIS datetime recognition by allowing the use of the 
     /// user's current timezone time as a relative datetime rather than a mix of absolute times and UTC times.
     /// </summary>
-    [MsGraphCustomActionRegistration(CheckAvailability.DeclarativeType)]
     public class CheckAvailability : Dialog
     {
+        [JsonProperty("$kind")]
         public const string DeclarativeType = "Microsoft.Graph.Calendar.CheckAvailability";
```
lauren-mills: This isn't using the same => syntax as all the others

skills/declarative/calendarSkill/runtime/customaction/MSGraphComponentRegistration.cs
```diff
@@ -40,6 +40,10 @@ public class MSGraphComponentRegistration : ComponentRegistration, IComponentDec
     {
         public IEnumerable<DeclarativeType> GetDeclarativeTypes(ResourceExplorer resourceExplorer)
         {
+            yield return new DeclarativeType<CheckAvailability>(CheckAvailability.DeclarativeType);
```
lauren-mills, line 43: Why are these three added individually rather than based on the attribute?

lauren-mills (changes requested):

skills/declarative/calendarSkill/runtime/customaction/MSGraphComponentRegistration.cs
thomas-chung: Because they aren't technically MsGraph custom action. I am thinking about keeping this with the custom actions for the calendar skill.

skills/declarative/calendarSkill/runtime/customaction/Actions/MSGraph/CheckAvailability.cs
thomas-chung: This is deliberately keeping the same as the old existing style. The new style is basically kept for MsGraph custom actions and this isn't an MsGraph custom action, it has more to do with transforming the result from graph and filtering things. Eventually when I break out the MsGraph custom action package this would stay with the calendar skill.

lauren-mills (approved):

PR #456, opened 2020-12-16
Experimental Runtime samples

tracyboehrer:


experimental/samples/runtime/dotnet/echo-bot/echo-bot/echo-bot.csproj
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.NET.Sdk.Web">
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <AspNetCoreHostingModel>OutOfProcess</AspNetCoreHostingModel>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="3.1.8" />
+    <PackageReference Include="Microsoft.Bot.Builder.Runtime" Version="4.12.0-daily.preview.20201207.192346.a48b032" />
+    <PackageReference Include="MSTest.TestFramework" Version="2.1.2" />
```
peterinnesmsft, line 10: I thought I saw the MSTest reference the other day and thought I was going crazy, but I guess not! Do we need this reference? I would assume not?

experimental/samples/runtime/dotnet/echo-bot/echo-bot/runtime.json
```diff
@@ -0,0 +1,17 @@
+{
```
peterinnesmsft, line 1: We should be able to update this to reflect the latest changes to single adapter, etc.

experimental/samples/runtime/dotnet/echo-bot/echo-bot/Dialogs/echo/generated/echo.lu.qna.dialog
```diff
@@ -0,0 +1,4 @@
+{
+    "$kind": "Microsoft.CrossTrainedRecognizerSet",
+    "recognizers": []
+}
```
peterinnesmsft, line 4: Do we expect these samples to be used primarily in Composer? If so, I would remove the generated folder and its contents from both and let it handle producing these directly.

experimental/samples/runtime/dotnet/slack-adapter/slack-adapter/Dialogs/echo/generated/echo.lu.qna.dialog
```diff
@@ -0,0 +1,4 @@
+{
+    "$kind": "Microsoft.CrossTrainedRecognizerSet",
+    "recognizers": []
+}
```
peterinnesmsft, line 4: Same thing here, if we intend to primarily use in Composer, I would remove this and let Composer handle generating it directly.

peterinnesmsft (approved):

experimental/samples/runtime/dotnet/echo-bot/echo-bot/echo-bot.csproj
tracyboehrer: Probably not for a sample?

experimental/samples/runtime/dotnet/echo-bot/echo-bot/runtime.json
tracyboehrer: Yes.  This is dependent on an older version of the Runtime package.  The templates and samples will have to change when the newer Runtime package is published.

experimental/samples/runtime/dotnet/echo-bot/echo-bot/Dialogs/echo/generated/echo.lu.qna.dialog
tracyboehrer: Another great question.  One avenue could be you could run the sample directly.  The other is that you use it in Composer.  So... Not sure at this point.

experimental/samples/runtime/dotnet/echo-bot/echo-bot.sln
```diff
@@ -0,0 +1,25 @@
+
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.30503.244
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "EchoBot", "echo-bot\echo-bot.csproj", "{21547831-73F3-439C-9927-5E7DCA9FD7EA}"
```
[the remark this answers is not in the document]
tracyboehrer: P.S.  I was 100% sure you'd ask this.

ryanlengel (approved):

PR #538, opened 2021-02-02
Move MsGraph custom action to netstandard2.0 by not using ToHashSet()

thomas-chung:
### Purpose

### Changes

### Tests

### Feature Plan

#### General

### DIFF
--- a/experimental/components/MsGraph/Actions/GetContacts.cs
+++ b/experimental/components/MsGraph/Actions/GetContacts.cs
@@ -90,7 +90,10 @@ namespace Microsoft.BotFramework.Composer.CustomAction.Actions.MSGraph
 
             if (people?.Count > 0)
             {
-                var existingResult = contactsResult.SelectMany(c => c.EmailAddresses).ToHashSet();
+                var result = contactsResult.SelectMany(c => c.EmailAddresses);
+
+                var existingResult = new HashSet<string>(result);
+
                 foreach (var person in people)
                 {
                     var emailAddresses = new List<string>();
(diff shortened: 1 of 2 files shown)


lauren-mills (approved):

experimental/components/MsGraph/Actions/GetContacts.cs
```diff
@@ -90,7 +90,10 @@ protected override async Task<List<CalendarSkillContactModel>> CallGraphServiceW
 
             if (people?.Count > 0)
             {
-                var existingResult = contactsResult.SelectMany(c => c.EmailAddresses).ToHashSet();
+                var result = contactsResult.SelectMany(c => c.EmailAddresses);
```
peterinnesmsft, line 93: I would just move the IEnumerable straight into the constructor.

experimental/components/MsGraph/Actions/GetContacts.cs
```diff
@@ -90,7 +90,10 @@ protected override async Task<List<CalendarSkillContactModel>> CallGraphServiceW
...
+
+                var existingResult = new HashSet<string>(result);
```
peterinnesmsft, line 95: Presumably this should use StringComparer.OrdinalIgnoreCase?

peterinnesmsft (changes requested):

experimental/components/MsGraph/Actions/GetContacts.cs
thomas-chung: The default behavior from before does not use ordinal ignore case. We might as well fix it.

experimental/components/MsGraph/Actions/GetContacts.cs
thomas-chung: Sure!

lauren-mills (approved):

PR #562, opened 2021-02-12
Adding bot template generator

pavolum:
### Purpose
The bot template generator added in this PR is a yeomen generator that generates a the boilerplate structure for a bot template using the component model (this under the hood is itself a yeomen generator). This is a tool intended to help template creators get started when creating bot templates for use in and outside of Composer. 

### Changes
What is generated?

- When the user uses this tool it will generate an empty bot generator with a few areas where they will be advised to add their content which is specifc to the template they are building (outlined in Readme)
- Documentation is added to the tools readMe to indicate how to use this tool and which fields they will need to customize with their template specific implementation.

**Outputted file structure for what this tool generates:**
image

### Shortcomings
This initial generator-generator suffices for current r12 template requirements however does have shortcomings and areas of improvement. These are starting to being tracked in these two issues.
https://github.com/microsoft/botframework-components/issues/564
https://github.com/microsoft/botframework-components/issues/563


pavolum:
Question: 
Right now  a template generated by this tool does not have a 'scripts/*' or 'recognizers/*' directory. When the generated template is used in the Composer creation flow then Composer will add these directories. 

In the next release we will have to ensure these are generated outside of Composer to account for non Composer creation scenarios. 

In the meantime should we leave this as is or have it bootstrap the 'scripts/' and 'generators/' folder as part of the template generation.

This also applies to the existing templates that this generator has already created (mainly conversational core). Should the conversational core template bootstrap 'scripts' and 'recognizers' or assume the user has created through Composer and assume composer has generated these directories.

This is a question only for the R12 release, post R12 we will ensure all these things are handled outside Composer.

tools/generator-bot-template-generator/app/templates/generators/app/templates/.gitignore
```diff
@@ -0,0 +1,2 @@
+# prevent appsettings.json get checked in
+**/appsettings.json
```
gabog, line 2: Not sure if I have the right context but appsettings.json should be OK to check in but we need to support appsettings.Development.json and have that one ignored for local testing.

pavolum: Good catch! I have removes that. This is the gitignore file that will be bootstrapped in the end project for the user. Its mainly empty as its up to them to what they want to ignore but they would presumable not want to gitignore appsettings. I have fixed this.

tools/generator-bot-template-generator/app/index.js
```diff
@@ -0,0 +1,64 @@
+'use strict';
+const path = require('path');
+const Generator = require('yeoman-generator');
+const askName = require('inquirer-npm-name');
+const _ = require('lodash');
+const extend = require('deep-extend');
+const mkdirp = require('mkdirp');
+
+function makeGeneratorName(name) {
+  name = _.kebabCase(name);
```
peterinnesmsft, line 10: lol kebab case

tools/generator-bot-template-generator/app/index.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  prompting() {
+    return askName(
+      {
+        name: 'name',
+        message: 'Your generator name',
+        filter: makeGeneratorName,
+        validate: str => {
+          return str.length > 'generator-'.length;
```
peterinnesmsft, line 27: Why does it need to be longer than 'generator-'? I'm not understanding the purpose of this at face value.

tools/generator-bot-template-generator/app/templates/generators/app/templates/emptyBot.botproj
```diff
@@ -0,0 +1,5 @@
+{
+  "$schema": "https://raw.githubusercontent.com/microsoft/BotFramework-Composer/main/Composer/packages/server/schemas/botproject.schema",
+  "name": "emptyBot",
```
peterinnesmsft, line 3: Should we be renaming these files from emptyBot? Or is that being handled elsewhere?

tools/generator-bot-template-generator/app/templates/generators/app/templates/language-generation/en-us/common.en-us.lg
```diff
@@ -0,0 +1,2 @@
+# WelcomeUser
+- Welcome to the EmptyBot sample
```
peterinnesmsft, line 2: Would be nice to tweak this to something that reflects the generated bot I guess. Not major.

tools/generator-bot-template-generator/yarn.lock
```diff
@@ -0,0 +1,8528 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
peterinnesmsft, line 1: Do we even need this? This is all pure Node, no?

tools/generator-bot-template-generator/package.json
```diff
@@ -0,0 +1,73 @@
...
+  "name": "generator-bot-template-generator",
+  "version": "0.0.1",
+  "description": "Generate a bot template generator for use with Microsoft Bot Framework",
+  "author": "Microsoft",
+  "files": [
+    "app"
+  ],
+  "keywords": [
+    "bot-template-generator",
+    "microsoft bot template"
```
peterinnesmsft, line 11: Should these not be separated out into "microsoft" "bot" "template"?

peterinnesmsft (changes requested):
Few outstanding questions/comments.

tools/generator-bot-template-generator/README.md
```diff
@@ -0,0 +1,76 @@
+# bot-template-generator
```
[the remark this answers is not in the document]
pavolum: Yeah thats the intention once we merge the initial version. It will likely only be used internally at first but having it publicly available will make it easier for internal devs to use it. I could see external devs using it as well down the line especially when we all custom feeds in the creation flow to populate the template selection screen and to get the open source community making third party templates post build. 

I was wondering the same thing as far as location, @ryanlengel what do you think, I believe we initially spoke about putting it in the tools directory but it is a generator so I could see the argument for both.

tools/generator-bot-template-generator/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Getting started
+
+### Starting with published version of generator-bot-template-generator 
+- Install: `npm install -g generator-bot-template-generator`
+- Run: `yo bot-template-generator`
+
+### Starting with local version of generator-bot-template-generator 
+- Navigate to the root dir of your local version of `generator-bot-template-generator` and run `npm install`
+- Navigate to the directory where you want to instantiate your bot template generator and run `yo {path to your local versions of generator-bot-template-generator's ./app/index.js file}`
```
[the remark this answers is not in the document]
pavolum: Ill add that, Im a fan of hitting it directly but its user preference really

tools/generator-bot-template-generator/yarn.lock
ryanlengel: https://classic.yarnpkg.com/blog/2016/11/24/lockfiles-for-all/

tools/generator-bot-template-generator/package.json
ryanlengel: Andrew has a task next week to go through and consolidate all of the package naming, keywords, etc.

tools/generator-bot-template-generator/README.md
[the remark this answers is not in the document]
ryanlengel: Agree, since this will be published this should move to /generators. /tools should be reserved for internal scripts we use to maintain the repo/manage technical debt

tools/generator-bot-template-generator/app/index.js
pavolum: For example, the user can enter either enter 'generator-conversational-core' or 'conversational-core', if they enter the ladder then the `filter` function will append 'generator-' as a prefix to 'conversational-core'. At the end of the day it needs that prefix and this allows the user to manually add it or not. 

The validate function just loosely ensures that 'generator-' is in the name by checking length.  its unlikely that this validate will ever be called and return false unless someone changes the `filter` function. I can remove the validate function entirely  for this reason or if desired to a stronger check to check that the name starts with 'generator-' as opposed to checking length.

tools/generator-bot-template-generator/app/templates/generators/app/templates/emptyBot.botproj
pavolum: That is being tracked here https://github.com/microsoft/botframework-components/issues/563

For R12, we are assuming creation within Composer and Composer handles the changing of end file names and this variable to the bot name the user chooses on creation.

tools/generator-bot-template-generator/package.json
pavolum: Handled

tools/generator-bot-template-generator/yarn.lock
pavolum: Not sure I understand, we need a yarn.lock file (or package.lock) for any solution that has dependencies right? There arent many dependencies but there are some. 

I did notice I had both yarn.lock and package-lock which is a mistake so I have removed yarn.lock in favor of npm's package.lock and also removed unused dependencies. Good catch, assuming this is what you meant :)

tools/generator-bot-template-generator/app/templates/generators/app/templates/language-generation/en-us/common.en-us.lg
pavolum: Yeah I wonder if we should even have any dialogs on initial template creation. This essentially outputs an empty-bot-generator and the user presumably will generally replace all the scaffolded empty bot .lg and .lu. I do like the idea of having something there so I will keep it as is. 

As far as templating the name there are issues when the generated generators template file has template variables because the generator generator does not have a definition for those variables in the copyTpl() function but I am going to try something clever. standby.

tools/generator-bot-template-generator/app/index.js
pavolum: Any change needed here or just amused? Its a simple little function. Could always write it ourselves but we already have the lodash dependency so may as well use it.

tools/generator-bot-template-generator/app/templates/generators/app/templates/language-generation/en-us/common.en-us.lg
pavolum: Resolved

tools/generator-bot-template-generator/README.md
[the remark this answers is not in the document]
pavolum: Okay I have moved it to the generators dir

peterinnesmsft:
Per our other renamings, should this now become @microsoft/generator-microsoft-bot-template-generator?

generators/generator-bot-template-generator/app/templates/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "<%= name %>",
+  "version": "0.0.1",
```
peterinnesmsft, line 3: I would just default to 1.0.0 for the file output by the generator.

generators/generator-bot-template-generator/app/templates/generators/app/templates/settings/appsettings.json
```diff
@@ -0,0 +1,57 @@
+{
```
peterinnesmsft, line 1: We will probably need to make changes here based on @carlosscastro's changes, so might be worth to quickly check and sync when you have a moment.

generators/generator-bot-template-generator/package.json
```diff
@@ -0,0 +1,68 @@
+{
+  "name": "generator-bot-template-generator",
+  "version": "0.0.1",
```
peterinnesmsft, line 3: Let's snap to 1.0.0 and use @ryanlengel's pipeline changes for the purposes of adding version suffix etc until we're comfortable to do a GA release.

peterinnesmsft (approved):
Couple of things to quickly check and sync before pushing, but overall looks good.

generators/generator-bot-template-generator/app/templates/package.json
pavolum: handled

generators/generator-bot-template-generator/package.json
pavolum: Changed version # and spoke with Ryan about pushing yaml file in a separate PR

generators/generator-bot-template-generator/app/templates/generators/app/templates/settings/appsettings.json
pavolum: Standardized appsettings handled

PR #568, opened 2021-02-13
Pavolum/generator empty bot

pavolum:
### Purpose
This PR adds the empty bot template generator which is a required template for the R12 preview release, this will also be the empty bot generator that will be used under the hood when the user selects the qna template, however the qna generator will route to the qna model once empty bot is created. 

### Changes
Pretty simple template with no component dependencies and just the boilerplate file structure. More complex bots can build off this template via adding packages from the package manager or adding dialogs through Composer


.gitignore
```diff
@@ -266,6 +266,7 @@ FakesAssemblies/
 /tools/**/node_modules/
 /skills/**/node_modules/
 /samples/**/node_modules/
+/generators/**/node_modules/
```
ryanlengel, line 269: One of my PRs replaced all this with just `**/node_modules` so we can catch all.

pavolum: Okay I will remove this then?

pavolum: Wait should I remove all of the node_modules gitignores then and have your PR handle this?

pavolum: Okay I have pulled your change and reverted mine

ryanlengel (approved):

PR #581, opened 2021-02-17
feat: generator-adaptive-bot plugin support

peterinnesmsft:
### Purpose
Merging in latest changes to generator-adaptive-bot to support plugins.

### Changes
- Adding support for pluginDefinitions to generator options to enable other generators to add one or more plugins out of the box when scaffolding the bot.
- Due to package publishing constraints being enforced for security, removing multiple package sources from template NuGet.config.

### Tests
Manually verified with Composer integration and validation.

PR #585, opened 2021-02-18
Generators: Update to latest runtime, update config, add adapter support, minor fixes

### LINKED ISSUE #583: Generators: configurable adapter support

### LINKED ISSUE #584: Generators: Update runtime config to runtime version in SDK main

carlosscastro:
Close #583 
Close #584 

### Purpose

Tactical changes to the generators to support multiple adapters and be up to date with the runtime progress in dotnet SDK.

### Changes

- Generator controllers: Support multiple adapters
- Runtime version update: New version is 4.12.0-rc0 / 4.12.0-rc0.preview
- Generator default settings now in line with SDK 4.12.0-rc0 / 4.12.0-rc0.preview runtime
- Add temporary workaround for this bug: https://github.com/microsoft/botbuilder-dotnet/issues/5239

### Tests

Manually tested adaptive bot, empty bot, multibot with calendar (did find a list of blocking things outside of this PR that I'll share separately)

### Feature Plan
Next comes to test this out, quickly fix bugs, integrate in to composer, test more.


generators/generator-adaptive-bot/generators/app/templates/assets/appsettings.json
```diff
@@ -1,7 +1,29 @@
...
+      "useInspection": false,
+      "traceTranscript": false
+    },
+    "telemetry": {
+      "instrumentationKey": "",
+      "logPersonalInformation": false,
+      "logActivities": true
+    },
+    "plugins": [
+      {
```
carlosscastro, line 18: I saw that we had a pluginDefinitions argument in the index.json but it didn't seem to be picked up anywhere. Doing this makes things just work, but wanted to check what was the original plan for the[ pluginDefinitions defined here](https://github.com/microsoft/botframework-components/search?q=plugindefinitions). Thoughts?

generators/generator-adaptive-bot/generators/app/templates/dotnet/functions/Startup.cs
```diff
@@ -11,21 +14,18 @@ public class Startup : FunctionsStartup
         public override void Configure(IFunctionsHostBuilder builder)
         {
             builder.Services.AddBotRuntime(builder.GetContext().Configuration);
+
+            // The workaround below will be removed next SDK patch, when this bug fix gets released: https://github.com/microsoft/botbuilder-dotnet/issues/5239
+            // In the meantime, we're guaranteed to have CoreAdapter registered as IBotFrameworkHttpAdapter, so look it up and register it as BotAdapter.
+            builder.Services.AddSingleton(sp => (BotAdapter)sp.GetServices<IBotFrameworkHttpAdapter>().First(a => typeof(BotAdapter).IsAssignableFrom(a.GetType())));
```
carlosscastro, line 20: I hate this! If there is an Rc1 it will go away, otherwise we'll ship a 4.12.1 right after 4.12 to remove this line.

generators/generator-adaptive-bot/generators/app/templates/dotnet/functions/Startup.cs
```diff
@@ -11,21 +14,27 @@ public class Startup : FunctionsStartup
...
 
-            configurationBuilder.ConfigurationBuilder.AddBotRuntimeConfiguration(
-                applicationRoot,
-                settingsDirectory,
-                isDevelopment);
+            configurationBuilder.ConfigurationBuilder.AddBotRuntimeConfiguration(applicationRoot, settingsDirectory);
+        }
+
+        private static BotAdapter RegisterCoreBotAdapter(IServiceCollection services)
+        {
```
carlosscastro, line 32: This cruft should go away as soon as we add a new RC build or we'll do a preview patch right after Rc0 ships.

generators/generator-adaptive-bot/generators/app/templates/assets/appsettings.json
peterinnesmsft: There are updates that still need to be checked in. I have an open PR out for these: https://github.com/microsoft/botframework-components/pull/581

generators/generator-adaptive-bot/generators/app/templates/dotnet/webapp/Startup.cs
```diff
@@ -39,5 +43,14 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
                    endpoints.MapControllers();
                });
         }
+
+        private static BotAdapter RegisterCoreBotAdapter(IServiceCollection services)
```
carlosscastro, line 47: This cruft should go away as soon as we add a new RC build or we'll do a preview patch right after Rc0 ships.

generators/generator-adaptive-bot/generators/app/templates/assets/appsettings.json
carlosscastro: Ahh that makes sense. So maybe I will remove them from the appsettings.json so the pluginsDefinition injects that setting which seems better.

generators/generator-adaptive-bot/generators/app/templates/assets/appsettings.json
```diff
@@ -1,7 +1,29 @@
 {
   "MicrosoftAppPassword": "",
   "MicrosoftAppId": "",
-  "applicationInsights": {
-    "InstrumentationKey": ""
+  "defaultLocale": "en-US",
+  "runtimeSettings": {
```
carlosscastro, line 5: @pavolum this section was added to all configs since this is the default minimum stuff. However if composer provides a 'base' setting, we could move this runtimeSettings node to composer and only keep 'additive' stuff in these appsettings.json. So it depends more on what composer does for us, how much it knows about our settings and what the design is on that regard.

In the long term, I'd love to get composer to know as little as possible about our config shape. The composer runtime plugin can know, but not he rest of composer ideally. Thoughts?

generators/generator-adaptive-bot/generators/app/templates/assets/appsettings.json
carlosscastro: discussed offline and removed plugin info

generators/generator-adaptive-bot/generators/app/templates/assets/appsettings.json
carlosscastro: Discussed with Patrick, it's fine as-is for the time being.

pavolum: Makes sense to me! At the moment Composer does not provide much of a base set of settings for the runtime obj which does lead to the template generators themselves declaring a similar set of boilerplate config in the runtime setting. 
As far as 'composer to know as little as possible about our config shape' that is the plan for the next release when we focus a bit more on non Composer environments like running F5 in VS.

carlosscastro:
Starting the 11-conflict-party

generators/generator-microsoft-bot-adaptive/generators/app/templates/dotnet/webapp/Controllers/BotController.cs
```diff
@@ -1,34 +1,63 @@
...
+                {
+                    _adapters.Add(settings.Route, adapter);
+                }
+            }
         }
 
         [HttpPost]
         [HttpGet]
-        public async Task PostAsync()
+        [Route("api/{route}")]
```
carlosscastro, line 41: Detail for the future: there are reserved words that route cannot take. Most of them enforced by asp.net core routing. Examples: `action`, `area`, `controller`, `handler`, `page`. On our side, `skills` is a reserved word as well. There is a point to be made about adding validations for these in composer where folks configure adapters. However that would happen in the runtime plugin so that composer doesn't have asp.net specific code. Furthermore, we could add another layer of validation here, forbidding registering adapters for those routes. Thoughts? Not a pri 0 or 1 but nice to have for R13.

generators/generator-microsoft-bot-adaptive/generators/app/templates/dotnet/functions/Triggers/MessagesTrigger.cs
```diff
@@ -34,21 +47,35 @@ public MessagesTrigger(IBotFrameworkHttpAdapter adapter, IBot bot)
...
+            if (string.IsNullOrEmpty(route))
             {
-                return new OkResult();
+                throw new ArgumentNullException(nameof(route));
             }
 
-            return new ContentResult()
+            IBotFrameworkHttpAdapter adapter;
+
+            if (_adapters.TryGetValue(route, out adapter))
```
peterinnesmsft, line 59: nit - collapse into:

`if (_adapters.TryGetValue(route, out IBotFrameworkHttpAdapter adapter))`

generators/generator-microsoft-bot-adaptive/generators/app/templates/dotnet/functions/Triggers/MessagesTrigger.cs
```diff
@@ -34,21 +47,35 @@ public MessagesTrigger(IBotFrameworkHttpAdapter adapter, IBot bot)
...
+                {
+                    return new OkResult();
+                }
+                
+                return new ContentResult()
+                {
+                    StatusCode = req.HttpContext.Response.StatusCode,
+                };
+            }
+            else
```
peterinnesmsft, line 75: I think we can remove the else statement since we will always return if we hit true on the above condition, and just move the KeyNotFoundException up a scope.

generators/generator-microsoft-bot-adaptive/generators/app/templates/dotnet/webapp/Controllers/BotController.cs
```diff
@@ -1,34 +1,63 @@
...
-            // The adapter will invoke the bot.
-            await this._adapter.ProcessAsync(Request, Response, _bot).ConfigureAwait(false);
+            if (string.IsNullOrEmpty(route))
+            {
+                throw new ArgumentNullException(nameof(route));
+            }
+
+            IBotFrameworkHttpAdapter adapter;
+
+            if (_adapters.TryGetValue(route, out adapter))
```
peterinnesmsft, line 51: Same comments here on inline out variable declaration and removal of else statement/inversion of condition.

generators/generator-microsoft-bot-adaptive/generators/app/templates/dotnet/webapp/Controllers/BotController.cs
peterinnesmsft: Agreed on this validation not being tied to ASP.NET specific implementation. We probably want to discuss validations all up - I think there's a broad category of improvements we should make around that area, and this would fit into that bucket in my mind.

generators/generator-multibot/generators/app/templates/settings/appsettings.json
```diff
@@ -1,30 +1,35 @@
 {
-  "feature": {
-    "UseShowTypingMiddleware": false,
-    "UseInspectionMiddleware": false,
-    "RemoveRecipientMention": false
-  },
+  "MicrosoftAppPassword": "",
```
peterinnesmsft, line 2: Don't think this is something we need to change in this PR, but I think seeing these changes is telling me we should find a way to centralize as much of this as possible into the generator-microsoft-bot-adaptive layer. I'll follow up on this later and see what we can do.

peterinnesmsft (approved):
Couple of minor tweaks but otherwise looks good.

PR #586, opened 2021-02-18
Updating package names to discussed naming conventions

pavolum:
### Purpose
Updating the package names and versions for generator-conversational-core, generator-adaptive-bot and generator-empty-bot to scoped Microsoft namespace per discussion with the larger team on desired naming for R12 release.


generators/generator-conversational-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "generator-conversational-core",
-  "version": "1.0.12",
+  "name": "@microsoft/generator-microsoft-bot-conversational-core",
+  "version": "1.0.0-preview1",
```
ryanlengel, line 3: You can make this one "preview.1" as well

peterinnesmsft: For my benefit, are we going with preview.1, or preview1? I don't mind either way, but I just changed them in the other components to preview1. I don't mind either way, but now would be the time to make the call and pull the trigger.

peterinnesmsft (changes requested):
Can you rename corresponding directory structures and whatnot to reflect updated package names?

pavolum:
starting a new PR to handle merge conflicts

PR #589, opened 2021-02-18
Pavolum/change generator names

pavolum:
### Purpose
- Changing package names and versions in package.json to those agreed upon by the team
- Updating dir names to reflect new package names


peterinnesmsft (approved):
I think this otherwise looks good. Let's hold off on publishing the preview versions of the packages immediately though - we should make sure these are aligned with the standard SDK package version format for npm.

PR #592, opened 2021-02-19
Configure prerelease version in pipelines

ryanlengel:
### Purpose
Prerelease packages that we publish should follow the SDK's format of dateStamp & commitHash

### Changes
#### Pipelines
- Added a variable `DeploymentRing` to existing pipelines to indicate if the component should have prerelease tags "alpha" or "preview", or if it is "stable" and generally available
- Added versioning-steps template to all pipelines to determine the prerelease version tag of {deploymentRing}.{dateStamp}.{commitHash}. This is stored in environment variable $(VersionSuffix) referenced in other steps. If the deploymentRing is "stable", no version suffix is saved.
#### Dotnet (.csproj)
- Components with a .csproj use the dotnet cli to create the package and these pick up the $(VersionSuffix) variable by default.
- Removed the `<VersionSuffix>` property in .csproj
image
image

#### Nuget (.nuspec)
- Components with a .nuspec use the nuget cli, added contiional steps:
  - Deployment ring is "stable", run `nuget pack` like before
  - Deployment ring is not "stable", run `nuget pack` with arguments to include version suffix
image
image

#### NPM
- Added conditional step if deployment ring is not "stable"
  - Run Powershell script to regex replace the package.json version appended with version suffix
image
image


build/yaml/templates/versioning-steps.yml
```diff
@@ -0,0 +1,15 @@
+steps:
+  - powershell: |
+      "Deployment Ring = $(DeploymentRing)";
+      $deploymentRing = "$(DeploymentRing)";
+      $dateStamp = (Get-Date -format "yyyyMMdd");
+      $commitHash = "$(Build.SourceVersion)".SubString(0,7);
+      $vs = $dateStamp + "." + $commitHash;
+
+      if ($deploymentRing -ne "stable") {
```
peterinnesmsft, line 9: Probably worth forcing a .ToLowerInvariant() on $deploymentRing to ensure consistent casing.

build/yaml/templates/versioning-steps.yml
```diff
@@ -0,0 +1,15 @@
+steps:
+  - powershell: |
+      "Deployment Ring = $(DeploymentRing)";
```
peterinnesmsft, line 3: Is this first line just for logging purposes?

build/yaml/templates/versioning-steps.yml
```diff
@@ -0,0 +1,15 @@
...
+      $deploymentRing = "$(DeploymentRing)";
+      $dateStamp = (Get-Date -format "yyyyMMdd");
+      $commitHash = "$(Build.SourceVersion)".SubString(0,7);
+      $vs = $dateStamp + "." + $commitHash;
+
+      if ($deploymentRing -ne "stable") {
+        $vs = $deploymentRing + "." + $vs;
+        Write-Host "##vso[task.setvariable variable=VersionSuffix;]$vs";
```
peterinnesmsft, line 11: Does Write-Host have any significance here? I'm assuming it does since the other ones look to probably be logging but don't have it.

build/yaml/templates/versioning-steps.yml
ryanlengel: Yes, helps to validate when the builds are running like so:
image

build/yaml/templates/npm-package-steps.yml
```diff
@@ -4,14 +4,31 @@ steps:
...
+    # Get existing version from package.json
+    $packageJson = Get-Content package.json;
+    $packageJsonData = $packageJson | ConvertFrom-Json;
+
+    "VERSION: $($packageJsonData.version)"
+    "SUFFIX: $(VersionSuffix)";
+    $version = $($packageJsonData.version) + "-" + "$(VersionSuffix)";
+
+    # Replace matched version
+    (Get-Content package.json) `
```
peterinnesmsft, line 17: Why not just update $packageJsonData.version directly, then ConvertTo-Json | Out-File? Seems a little simpler than regex matching.

build/yaml/templates/component-template.yml
```diff
@@ -15,6 +15,7 @@ stages:
         - job: Build
           displayName: Build and package
           steps:
+          - template: versioning-steps.yml
           - template: dotnet-package-steps.yml
```
peterinnesmsft, line 19: I was expecting to see some kind of corresponding change to dotnet-package-steps.yml similar to how you've done for npm and nuspec. Is it not needed?

build/yaml/templates/versioning-steps.yml
peterinnesmsft: Excellent!

peterinnesmsft (changes requested):
Looks like there should be a change to dotnet-package-steps.yml - waiting for confirmation. Otherwise looks good to me!

build/yaml/templates/versioning-steps.yml
ryanlengel: Yeah it is different, it's how PS can set environment variables back into ADO.

ryanlengel: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=classic%2Cpowershell#set-a-job-scoped-variable-from-a-script-1

build/yaml/templates/npm-package-steps.yml
ryanlengel: Sure that would be simpler but that's not exactly what one thinks of working after 10PM :P

build/yaml/templates/component-template.yml
ryanlengel: Nope! MSBuild picks up the $(VersionSuffix) environment variable on its own. I thought that I'd need to explicitly include it in packing steps but it worked automatically once the variable was configured.

thomas-chung (approved):
Signing off on the calendar package changes

build/yaml/templates/npm-package-steps.yml
peterinnesmsft: I mean, the real question is why you're working after 10:00pm, but I'll allow it :P lol

build/yaml/templates/component-template.yml
peterinnesmsft: Ahh, wonderful! Kinda magic but works for me!

peterinnesmsft (approved):

PR #596, opened 2021-02-19
feat: Adding plugin definitions to appsettings.json

peterinnesmsft:
### Purpose
Per latest changes to appsettings.json format with the runtime, we now need to place plugin definitions within appsettings.json instead of runtime.json, which has now been removed. Ideally the base generator-microsoft-adaptive-bot is fully responsible for this, but the multibot and calendar generators need to have these definitions manually added in the interim since they don't use the file handled by the base generator.

### Changes
- Re-incorporated improvements to _validate functions in generator-microsoft-adaptive-bot.
- Add _writeApplicationSettings to generator-microsoft-adaptive-bot to handle writing of plugin definition information to runtime settings.
- Updated generator-multibot and generator-preview-calendar-skill to include settingsPrefix in pluginDefinitions, as well as copy directly into custom appsettings.json to ensure the plugins are loaded out of the box correctly per latest changes.

### Tests
Manual validation.


generators/generator-microsoft-bot-adaptive/generators/app/index.js
```diff
@@ -179,4 +222,20 @@ module.exports = class extends Generator {
...
     }
+
+    _writeApplicationSettings() {
+        const botName = this.options.botName;
+        const fileName = 'appsettings.json';
+
+        const appSettings = this.fs.readJSON(this.templatePath(path.join('assets', fileName)));
+
+        for (const pluginDefinition of this.pluginDefinitions) {
+            appSettings.plugins.push(pluginDefinition);
```
carlosscastro, line 233: Should this be

```suggestion
            appSettings.runtimeSettings.plugins.push(pluginDefinition);
```

peterinnesmsft: Yes - good catch.

carlosscastro (approved):

PR #598, opened 2021-02-20
feat: Generator build YAML updates and dependency changes

peterinnesmsft:
### Purpose
Generator build YAML files need to be updated to point to renamed generators, and also modify slightly the version suffix to follow pattern set by other SDK packages consistently. Also updating package dependencies for generator-microsoft-bot-conversational-core and generator-microsoft-bot-empty to use generator-microsoft-bot-adaptive.

### Changes
See above.

### Tests
Flying by the seat of my pants to be honest.


carlosscastro (commented):
:shipit:

PR #623, opened 2021-02-23
feat: Remove clear element from scaffolded NuGet.config

peterinnesmsft:
### Purpose
We need to make sure we do not clobber other configured package sources for restoring packages. Due to security enforcements, the NuGet.config was updated to limit to a single source - we need to adjust this for the actual generated project itself and separate from our own build pipeline requirements.

### Changes
- Update handling of outputting NuGet.config to clear out 'clear' element from packageSources during scaffolding.

### Tests
Manually tested.


carlosscastro (approved):

carlosscastro:
We might have to revisit this, in that we're removing the text from a template we created, so it might be hard to spot for future developers

Issue #480, opened 2021-01-11
DevOps for the -component repository.


clearab:
## Description
Templates and packages (C# and JS) need an automated publishing pipeline to reduce friction for component developers.

## Work items
- [x] Pipeline steps
  - [x] dotnetBuildSteps - _Run `dotnet build`_
  - [x] dotnetPackSteps - _Run `dotnet pack`_
  - [x] dotnetPushSteps - _Run `dotnet push`_
  - [x] dotnetRestoreSteps - _Run `dotnet restore`_
  - [x] npmPublishSteps - _Run `npm publish`_
  - [x] nugetPackSteps - _Run `nuget pack` (for .nuspec)_
  - [x] nugetPushSteps - _Run `nuget push` (for .nuspec)_
  - [x] useDotnetSteps - _Use dotnet 2.x/3.x_
  - [x] codeSignSteps - _Sign nuget packages_
  - [x] versionTaggingSteps - _Add version tag based on deployment ring and branch_ 
  - [x] publishBuildArtifactsSteps - _Publish files to build aritfacts_
  - [x] debugSteps - Display environment variables and workspace_
- [x] Determine permissions to publish to nuget.org/npm GA
- [x] Determine Azure Artifacts feed and views for preview packages
- [x] Transition to SDK_v4 pipelines
 
### componentPipelineTemplate
_User can create a new template extending the componentPipelienTemplate, specifying a component type that calls related subtemplate._
#### Parameters
```
name: componentType
  displayName: 'What is the component type? Determines deployment steps.'
  type: string
  values:
  - declarativeAsset
  - plugin
  - generator

name: deploymentRing
  displayName: 'What is the component type? Determines deployment steps.'
  type: string
  default: privatePreview
  values:
  - privatePreview
  - publicPreview
``` 

### declarativeAssetPipelineStages
Declarative assets are published to both NuGet and NPM. This assumes they already include .nuspec and package.json files, respectively. Uses `nuget` rather than `dotnet` commands due to supportability of .nuspec file.

- Validate that required packaging files are provided. 
- Run `nuget pack`  
- Sign with the ESRP CodeSign task. 
- Publish .nupkg to build artifacts. 
- Publish node package to build artifacts. 

### generatorPipelineStages
Template generators are published to NPM. This assumes they already include a package.json file.

- Publish node packages to build artifacts. 

### pluginPipelineStages
[...]

PR #628, opened 2021-02-24
Pipelines: Add tag builds, versioning setup stage

ryanlengel:
### Purpose
Tagging builds with generated package versions for #480 

### Changes
- Replaced the componentType parameter with a ComponentType variable instead for all pipelines. Parameters cannot be passed into referenced templates while variables _can_ and shared across environments. This allows us to set condition properties per each stage/job in `component-template.yml` rather than evaluating if statements like previously done prior.
- Setup stage
  - Added nuget versioning steps that inspect the version provided in either .nuspec or .csproj file, and concatenate with calculated version suffix. Build is tagged with `NugetPackageVersion` and said variable is outputted to be picked up in the Build_pack stage.
  - Added npm versioning steps that inspect the version provided in package.json and concatenate with calculated version suffix. Build is tagged with `NpmPackageVersion` and said variable is outputted to be picked up in the Build_pack stage.
- Build_pack stage
  - dotnet build steps use the NugetPackageVersion variable when generating package via CLI
  - nuspec build steps use the NugetPackageVersion variable when generating package via CLI
  - npm build steps can use the NpmPackageVersion variable when generating package after updating package.json
 
#### Generator
image
image

#### Declarative assets
image
image

#### Code extension
image
image

#### General


build/yaml/templates/npm-package-steps.yml
```diff
@@ -4,25 +4,32 @@ steps:
...
+
 - powershell: |
-    # Get existing version from package.json
-    $packageJson = Get-Content package.json;
-    $packageJsonData = $packageJson | ConvertFrom-Json;
+    if (Test-Path -Path package.json) {
+      # Get package.json ojbect
+      $packageJson = Get-Content package.json;
+      $packageJsonData = $packageJson | ConvertFrom-Json;
+      $packageJsonDataVersion = $packageJsonData.version;
```
peterinnesmsft, line 17: It doesn't look like you're using $packageJsonDataVersion - remove?

peterinnesmsft (approved):

PR #726, opened 2021-03-10
Various bug fixes for Who Skill

thomas-chung:
### Purpose
- Fix bugs around various scenarios in Who skill
- Fix messaging to show the person's name when the have no peers in the org
- Fix csproj build failure when the runtime package was updated
- Handle when the user is not found in the director
- Update Find Person dialog to reflect design

### Changes

### Tests

### Feature Plan

#### General

### DIFF
--- a/skills/declarative/whoSkill/whoSkill/dialogs/FindPersonDialog/FindPersonDialog.dialog
+++ b/skills/declarative/whoSkill/whoSkill/dialogs/FindPersonDialog/FindPersonDialog.dialog
@@ -35,7 +35,9 @@
           "dialog": "UserListDialog",
           "options": {
             "UserList": "=$Result",
-            "AutoChooseFirst": "true"
+            "AutoChooseFirst": "true",
+            "Scenario": "PersonSearch",
+            "User": "=$nameEntity"
           },
           "resultProperty": "$SelectedUser"
         },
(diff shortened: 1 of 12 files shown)


skills/declarative/whoSkill/whoSkill/dialogs/FindPersonDialog/FindPersonDialog.dialog
```diff
@@ -23,7 +23,8 @@
           },
           "resultProperty": "$Result",
           "token": "=user.token.token",
-          "nameToSearchFor": "=$nameEntity"
+          "nameToSearchFor": "=$nameEntity",
+          "maxCount": 16
```
lauren-mills, line 27: 16 feels like an odd number here... Just want to make sure this is right

lauren-mills (approved):

skills/declarative/whoSkill/whoSkill/dialogs/FindPersonDialog/FindPersonDialog.dialog
thomas-chung: Yeah it is. The idea here is that if we get more than 15 then we just also return the "There are too many matches, please specify scenario". We don't need to get the exact count on the UI, just knowing it's got more than 15 is enough.

PR #747, opened 2021-03-16
Split HelpAndCancel in two separate dialogs that start with beginDial…

### LINKED ISSUE #681: HelpAndCancel: Split the Help and Cancel triggers/dialogs
image

The template has a single trigger for HelpAndCancel, which forwards to a HelpAndCancel dialog which triggers appropriately. Instead, these triggers should each have their own dialog and separate trigger in the root bot.

gabog:
Note, making this a draft PR so I can update it if [this](https://github.com/microsoft/BotFramework-Composer/issues/6417) composer issue gets fixed soon. 

Fixes #681

In this PR:

- Split HelpAndCancel in two separate dialogs that start with beginDialog (dialogs are in the same folder for now until we fix an issue in package manager).
- Deleted unnecessary lu files (the root dialog will have to do lu and call the help or cancel dialog when it detects the user intent). 
- Added cancel confirmation prompt to CancelDialog.
- Updated package description.

image


ryanlengel:
We need an agreement on what type of content these package should have and what installation paths we're supporting. We have two scenarios:

1. Install packages as part of a template in the creation flow
_As a bot dev I make a bot using one of the templates available in creation flow. Composer copies over the bot template and then installs dependent package, in this case copying over HelpAndCancel as separate dialogs/LG content. The LU triggers were manually hooked up in the root dialog by the template author prior. Once my bot has been created, I hit "Start Bot" and can validate it's working right away._
2. Install packages afterwards via package manager
_As a bot dev I make an empty bot available in creation flow. Composer copies over the bot template. I go to Package Manager and install HelpAndCancel myself. This copies over the separate dialogs/LG content, but has no LU triggers. Instead, it is up to myself to determine what triggers should go to my package._

For the 2nd scenario, wouldn't it be more valuable that we provide sample LU models to get started on? I imagine that most developers will change the LG content to fit their own bot anyways. On the other hand, I do see how it would be annoying to have packages installed in a template with content that you aren't even triggering in your bot.

ryanlengel (approved):
These look good, I'll add an issue in the SDK repo to track best practices on packages like we discussed

PR #754, opened 2021-03-17
feat: Move Welcome package content to BeginDialog and rename to WelcomeDialog

### LINKED ISSUE #684: Welcome: Use BeginDialog instead of ConversationUpdate
Hooking this package up in a root dialog will use ConversationUpdate, so it can just be a BeginDialog in the actual package.

Instead of this
image

Use this
image

ryanlengel:
### Purpose
Close #684 

### Changes
This PR addresses the feedback discovered during the previous bug bash on the Welcome package, namely:
- Move content from `OnConversationUpdate` to `OnBeginDialog`, it is up to the templates using this package to connect this to a Greeting trigger in the root bot.
- Rename to WelcomeDialog

image


packages/Welcome/exported/WelcomeDialog/language-generation/en-us/WelcomeDialog.en-us.lg
```diff
@@ -1,10 +1,10 @@
-# SendActivity_wStAtd()
+# SendActivity_PsAJ3p()
```
gabog, line 1: I think we are missing `[import](common.lg)` at the beginning
```suggestion
[import](common.lg)

# SendActivity_PsAJ3p()
```

gabog, line 1: We seem to be using the old form for SendActivity here:

image

As opposed to response variations:

image

which will generate an extra function:

image


I think we should adopt the new response editor form in our packages and templates

gabog (changes requested):
We are missing the default folders and files for:
* knowledge-base
* language-understanding
* recognizers

Composer seems to be expecting these or we will run into [this issue](https://github.com/microsoft/botframework-components/issues/755)

gabog (approved):

PR #761, opened 2021-03-17
Basic Who Skill Tests

thomas-chung:
### Purpose
Basic tests for Who skill

### Changes
- Refactor common core test base between Who and Calendar tests
- Who tests
- Bug fixes for the who skills in the graph custom actions

### Tests
It's covered by newly added tests

### Feature Plan

#### General


packages/Graph/MSGraphClient.cs
```diff
@@ -14,14 +14,33 @@ namespace Microsoft.Bot.Components.Graph
...
     public class MSGraphClient
     {
+#if DEBUG
+        private static IGraphServiceClient clientOverride;
+
+        /// <summary>
+        /// Allows setting test instance of the graph service client
+        /// </summary>
+        /// <param name="client">The client that should be used</param>
+        public static void RegisterTestInstance(IGraphServiceClient client)
```
lauren-mills, line 24: Why did you have to add this? Were the tests not able to auto resolve the httpclient? I was under the impression that was the case

thomas-chung: They can. However, I wanted to actually isolate the code at the IGraphServiceClient layer rather than the lower HttpClient layer which depends on hand-crafting and updating the json, which can be more error prone.

.github/workflows/build-and-test.yml
```diff
@@ -31,7 +31,12 @@ jobs:
         run:  dotnet build
         working-directory: ./skills/declarative/tests
 
-      # Run the tests
-      - name: Run tests
+      # Run the calendar tests
```
peterinnesmsft, line 34: While you're touching this, can you please update to only run the tests when needed? For example, I don't think there is any need to run tests if the change only touches generators for example.

lauren-mills (approved):

.github/workflows/build-and-test.yml
thomas-chung: Sure, we can add a directory filter on the actions.

PR #762, opened 2021-03-18
Fixes cancel intent and other conversational-core updates

### LINKED ISSUE #683: Conversation Core: Cancel intent picks up unknown utterances
The LU file needs to be balanced correctly, as it is the Cancel intent picks up all of the unknown utterances
image

gabog:
Fixes #683 

## Changes summary
* Added missing using for Microsoft.Extensions.Hosting in adaptive startup.cs
* Tuned up the LUIS model for convcore by removing some utterances and adding some utterances to the none
* Updated convcore references to use the latest welcome and help and cancel packages and wired up the dialogs to the root dialog
* Removed extra #designer element in conversational-core.dialog
* Removed reference to SDK schema from conversational-core.dialog (this is not needed)
* Added condition to the cancel intent to only trigger on 90% certainty. 
* Updated SendActivity calls to use the Response Editor format

**Note:** the luis model has too few utterances and it has no real end user intents, we should let the devs know if the documentation about the need to tune the LUIS model and make sure the know about the none intent too. 

Here are a screenshots of the resulting project:

## Root dialog
image
 
## Cancel intent with score condition
image

## Added none intent for training:
image

### DIFF
--- a/generators/generator-microsoft-bot-conversational-core/generators/app/templates/conversational-core.dialog
+++ b/generators/generator-microsoft-bot-conversational-core/generators/app/templates/conversational-core.dialog
@@ -1,11 +1,9 @@
 {
   "$kind": "Microsoft.AdaptiveDialog",
   "$designer": {
-    "$designer": {
       "name": "conversational-core",
       "description": "",
-      "id": "GnxrCe"
-    }
+      "id": "4pM5gc"
   },
   "autoEndDialog": true,
   "defaultResultProperty": "dialog.result",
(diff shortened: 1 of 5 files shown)


generators/generator-microsoft-bot-conversational-core/generators/app/templates/conversational-core.dialog
```diff
@@ -17,50 +15,88 @@
...
-            "id": "z0X16k"
+            "id": "IQMEuO"
           },
-          "activity": "${SendActivity_z0X16k()}"
+          "activity": "${SendActivity_IQMEuO()}"
         }
       ]
     }
   ],
-  "$schema": "https://raw.githubusercontent.com/microsoft/BotFramework-Composer/stable/Composer/packages/server/schemas/sdk.schema",
```
peterinnesmsft, line 63: Do we not still need a $schema value?

peterinnesmsft (approved):

generators/generator-microsoft-bot-conversational-core/generators/app/templates/conversational-core.dialog
gabog: Not according to @tomlm that was needed before we had a local schema.

ryanlengel (commented):
👍 will port these over to Assistant Core

PR #795, opened 2021-03-25
More Generator fixes

### LINKED ISSUE #787: QnA Region setting is missing from a new Conversational Core bot
## Version

Composer Version:  1.4.0-rc4-nightly.226081.b0a126

## Describe the bug
QnAMaker region setting is missing from a new conversational core bot, causing issues when trying to use QnA Maker.

image

## To Reproduce

1. Create a new Conversational Core bot with the preview experience in Composer
2. Once the bot is created go to settings and enable advanced to see the JSON
3. QnARegion setting is missing, which can cause an error when using QnA Maker.

## Expected behavior

QnARegion setting should be present (defaulted to westus)

Also raised an issue to ensure this setting is shown in the settings form UX as we do for LUIS - https://github.com/microsoft/BotFramework-Composer/issues/6535

peterinnesmsft:
### Purpose
Fixes #787.

This change addresses a number of small fixes that need to be made for R13.

### Changes
- NuGet.config removed from template files. This was originally added to support being able to restore packages from MyGet during development. This is no longer needed, and we don't need to expose the MyGet endpoint to consumers.
- Updating the adaptive generator to include dotnet template settings for including the .sln file, with default to true. This is to support the Enterprise Assistant template, which for now will need to override the default behavior so as to avoid generating separate .sln files for each bot.
- Adding 'region' property with default value of 'westus' to qna settings in appsettings.json.

### Tests
Manually verified.

### DIFF
--- a/generators/generator-bot-adaptive/generators/app/index.js
+++ b/generators/generator-bot-adaptive/generators/app/index.js
@@ -11,6 +11,9 @@ const options = rt.Record({
   botProjectSettings: rt
     .Record({ skills: rt.Dictionary(rt.Unknown) })
     .asPartial(),
+  dotnetSettings: rt.Record({
+    includeSolutionFile: rt.Boolean.Or(rt.Undefined),
+  }),
   modifyApplicationSettings: rt.Function,
   packageReferences: rt.Array(
     rt.Record({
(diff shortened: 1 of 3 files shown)


peterinnesmsft:
Please do not merge - need to confirm with @garypretty that the qna.region property is in fact the correct property name first!

joshgummersall (approved):

PR #803, opened 2021-03-26
Adds error handler to conversational core

### LINKED ISSUE #689: Conversational core template should implement default error handling in root dialog
I think it would be useful to provide a default error handling implementation in the conversational core template (and all templates derived from it)

I am not sure what the recommended practices are to implement error handling and logging in composer, but we should follow them if they exist.

The error handling should:
1. Display a friendly messagge giving a reason for the failure.
2. Send a trace event with detailed error info (Trace Activities only show in emulator).
3. Log the error using whatever logging mechanism is available in the underlying framework (if this is not done by the runtime).
4. Optionally, we could look into returning detailed error information if the bot is running in emulator (or have a global debug property set but we may need to think a bit more about this)  

## Runtime considerations:
We should check the adapter code in the runtime but it is important to note that we need additional logic to handle error that may happen in skills. If a bot is a skill consumer we should take a look at the error handling logic in [AdapterWithErrorHandle](https://github.com/microsoft/BotBuilder-Samples/blob/main/samples/csharp_dotnetcore/81.skills-skilldialog/DialogRootBot/AdapterWithErrorHandler.cs). If the bot is running as a skill, then we should make sure we support the logic that is implemented in [SkillAdapterWithErrorHandler](https://github.com/microsoft/BotBuilder-Samples/blob/main/samples/csharp_dotnetcore/81.skills-skilldialog/DialogSkillBot/SkillAdapterWithErrorHandler.cs)

gabog:
Fixes #689

This PR adds an error trigger to the root dialog and sends a friendly message and a trace with exception details to emulator (it is good to show devs where to find the exception details in memory).

It also removes the $schema property that was erroneously added to botName.dialog and removes a duplicated element from that dialog.

## Error Occurred action for telemetry

image

## Error Occurred action for emulator trace

image

## Emulator output when an error occurs

image

## Error in telemetry

image

### DIFF
--- a/generators/generator-bot-conversational-core/generators/app/templates/language-generation/en-us/botName.en-us.lg
+++ b/generators/generator-bot-conversational-core/generators/app/templates/language-generation/en-us/botName.en-us.lg
@@ -11,3 +11,12 @@
 - I'm not sure I understand. Can you please try again?
 - Hmm, I don't understand. Can you try to ask me in a different way. 
 - I didn't get that. Would you mind rephrasing and try it again.
+# SendActivity_2outgQ()
+[Activity
+    Text = ${SendActivity_2outgQ_text()}
+]
+
+# SendActivity_2outgQ_text()
+- Oops, looks like I'm stuck. Can you try to ask me in a different way?
+- Looks like I'm all mixed up. Let's try asking again, but maybe rephrase your request?
+- Sorry, it looks like something went wrong. Can you please try again?
--- a/generators/generator-bot-conversational-core/generators/app/templates/botName.dialog
+++ b/generators/generator-bot-conversational-core/generators/app/templates/botName.dialog
@@ -1,7 +1,6 @@
 {
   "$kind": "Microsoft.AdaptiveDialog",
   "$designer": {
-    "$designer": {
       "name": "<%= botName %>",
       "description": "",
       "id": "4pM5gc"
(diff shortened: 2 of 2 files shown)


ryanlengel:
@gabog the adaptive runtime OnTurnError handler logs the exception to a telemetry provider, did you get to test if that is still triggered here? If it is we are ok, if not then we need to add a telemetry track event in this dialog.

generators/generator-bot-conversational-core/generators/app/templates/language-generation/en-us/botName.en-us.lg
```diff
@@ -10,4 +10,13 @@
 - Sorry, I didn't get that
 - I'm not sure I understand. Can you please try again?
 - Hmm, I don't understand. Can you try to ask me in a different way. 
-- I didn't get that. Would you mind rephrasing and try it again.
\ No newline at end of file
+- I didn't get that. Would you mind rephrasing and try it again.
```
peterinnesmsft, line 13: nit: new line after

peterinnesmsft (approved):

generators/generator-bot-conversational-core/generators/app/templates/language-generation/en-us/botName.en-us.lg
gabog: This is as generated by composer, I can add it but if we update the template we may end up with the same if we just copy the lg.

gabog:
Hi @ryanlengel, based on what I see in the [DialogManager](https://github.com/microsoft/botbuilder-dotnet/blob/4.12/libraries/Microsoft.Bot.Builder.Dialogs/DialogManager.cs#L235) source if someone catches the error emitted there the exception will be swallowed and the adapter error handle will never get called (this will be the behavior in R13 as well, John and Tracy are looking at some other options too but not sure if they will make it).

ryanlengel (approved):

PR #808, opened 2021-03-26
Components: Update tags to new nomenclature

carlosscastro:


packages/HelpAndCancel/Microsoft.Bot.Components.HelpAndCancel.nuspec
```diff
@@ -11,7 +11,7 @@
         <projectUrl>https://github.com/Microsoft/botframework-components</projectUrl>
         <releaseNotes>First version of HelpAndCancel package</releaseNotes>
         <description>Dialogs to provide help and cancel utterance handling</description>
-        <tags>bf-component conversationalcore helpintent cancelintent preview</tags>
+        <tags>msbot-component msbot-content conversationalcore helpintent cancelintent preview</tags>
```
peterinnesmsft, line 14: Looking through the second order tags defined, msbot-content seems pretty generic in comparison to the others. Is there a reason we landed on this versus something like msbot-dialog?

peterinnesmsft, line 14: I think we should probably remove and standardize the other tags that are here while we are at it. For example, I don't think we should tag with conversationalcore or preview; we should probably add Microsoft Bot Framework tags, etc.

peterinnesmsft (approved):

PR #813, opened 2021-03-29
RC Build Pipeline Support

peterinnesmsft:
### Purpose
As part of our release process for our generator and component packages, we will need to easily cut RC builds of our packages for publishing. This change adds support to our build pipeline to do so.

### Changes
- Update pipeline files to reference ADO-defined variables 'DeploymentRingOverride' and 'ReleaseCandidateNumber' - the former enables overriding the DeploymentRing variable defined in the pipeline YAML (setting DeploymentRing in the ADO pipeline doesn't work, as YAML always takes precedence for variables); the latter is simply the numeric portion of the RC version suffix (e.g. rc0).
- 'rc' is referenced as a valid deployment ring value in our npm-versioning-steps.yml and nuget-versioning-steps.yml.
- Following behavior is updated in versioning logic:
  - If DeploymentRingOverride is defined, we use that over DeploymentRing.
  - If DeploymentRingOverride is 'rc', we use the ReleaseCandidateNumber as the version suffix; otherwise we fall back to the current behavior.
  - The full version suffix is then computed as a concat of the deployment ring and version suffix from above (the '.'/'-' separator is now baked into the prior calculated suffix for non-rc deployment rings).

On the Azure DevOps side, the expectation is that we will then just create new Release pipeline definitions for each component and package, so that we can very easily and quickly generate a new release version on demand when needed. This pipeline will specify a DeploymentRingOverride value of 'rc', and allow the specifying of the ReleaseCandidateNumber variable when scheduling (defaults to 0).

### Tests
Manually verified with


ryanlengel (approved):

tracyboehrer (approved):

tracyboehrer:
@peterinnesmsft We removed the Orchestrator yml since it's now in SDK

PR #822, opened 2021-03-30
Updated help and welcome and cancel packages to the latest versions.

gabog:
Updated references to latest packages.

This PR also fixes a style cop violation for sorting using statements in botcontroller and adds some code style changes like removing this. and using var instead of explicit types.


peterinnesmsft (changes requested):
Let me get the RC0 versions published and then we can roll with those.

peterinnesmsft (approved):

PR #850, opened 2021-04-05
[Enterprise assistant] Removed duplicated chooseIntent trigger

lauren-mills:
There is a duplicated trigger in the enterprise assistant template


ryanlengel (changes requested):
This implementation in core assistant uses 
```
"$designer": {
  "id": "YDQnEj"
}
```

Let's use the same to simplify future diffs

lauren-mills:
@ryanlengel the one that is left also uses that id, an exact copy just got added through a merge issue

ryanlengel (approved):

PR #853, opened 2021-04-06
Limit Graph Object Fields using Select() in People Skill

thomas-chung:
### Purpose
Limit the amount of data being retrieved in the People Skill-related Graph calls. This serves two purposes:
1. Limit the size of the payload we have to transfer across the wire
2. Limit unnecessary PII data exposure on behalf of end user

This change also updates the People skill to use the RC2 style bot app initialization.

### Changes
Switch to using explicit SELECT() clause in the Graph API to return only relevant information.

### Tests
Existing unit tests

### Feature Plan

Ideally the "select()" statement's list of columns are configurable. I think there could be some more thoughts put into it, like we can probably ask the bot custom action to add a parameter for "fields" required. That can be the next step but this is bare minimum.

#### General

### DIFF
--- a/packages/Graph/Actions/BaseMsGraphCustomAction.cs
+++ b/packages/Graph/Actions/BaseMsGraphCustomAction.cs
@@ -23,6 +23,11 @@ namespace Microsoft.Bot.Component.Graph.Actions
     public abstract class BaseMsGraphCustomAction : Dialog
     {
         /// <summary>
+        /// Name of the id field that Graph uses to identify objects in the system.
+        /// </summary>
+        public const string DefaultIdField = "id";
+
+        /// <summary>
         /// Initializes a new instance of the <see cref="BaseMsGraphCustomAction"/> class.
         /// </summary>
         /// <param name="callerPath">Caller path.</param>
--- a/packages/Graph/Actions/User/GetDirectReports.cs
+++ b/packages/Graph/Actions/User/GetDirectReports.cs
@@ -51,7 +51,11 @@ namespace Microsoft.Bot.Component.Graph.Actions
             int maxCount = (int)parameters["MaxResults"];
 
             // Create the request first then apply the Top() after
-            IUserDirectReportsCollectionWithReferencesRequest request = client.Users[userId].DirectReports.Request();
+            // TODO: Make this SELECT() clause configurable for different column names
+            // EXPLAINER: The reason we are limiting the field is for two reasons:
+            //            1. Limit the size of the graph object payload needed to be transferred over the wire
+            //            2. Reduces the exposure of end user PII (Personally Identifiable Information) in our system for privacy reasons. It's generally good practice to use what you need.
+            IUserDirectReportsCollectionWithReferencesRequest request = client.Users[userId].DirectReports.Request().Select("id,displayName,mail,businessPhones,department,jobTitle,officeLocation");
 
             if (maxCount > 0)
             {
--- a/packages/Graph/Actions/User/GetDirectReports.cs
+++ b/packages/Graph/Actions/User/GetDirectReports.cs
@@ -44,8 +44,8 @@ namespace Microsoft.Bot.Component.Graph.Actions
         /// <summary>
         /// Gets or sets the fields to select from the Graph API.
         /// </summary>
-        [JsonProperty("FieldsToSelect")]
-        public StringExpression FieldsToSelect { get; set; }
+        [JsonProperty("PropertiesToSelect")]
+        public ArrayExpression<string> PropertiesToSelect { get; set; }
 
         /// <inheritdoc/>
         public override string DeclarativeType => GetDirectReports.GetDirectReportsDeclarativeType;
@@ -55,10 +55,10 @@ namespace Microsoft.Bot.Component.Graph.Actions
         {
             string userId = (string)parameters["UserId"];
             int maxCount = (int)parameters["MaxResults"];
-            string fieldsToSelect = (string)parameters["FieldsToSelect"];
+            string propertiesToSelect = (string)parameters["PropertiesToSelect"];
 
             // Create the request first then apply the Top() after
-            IUserDirectReportsCollectionWithReferencesRequest request = client.Users[userId].DirectReports.Request().Select(fieldsToSelect);
+            IUserDirectReportsCollectionWithReferencesRequest request = client.Users[userId].DirectReports.Request().Select(propertiesToSelect);
 
             if (maxCount > 0)
             {
@@ -88,18 +88,23 @@ namespace Microsoft.Bot.Component.Graph.Actions
             }
 
             // Select minimum of the "id" field from the object
-            string fieldsToSelect = DefaultIdField;
+            string propertiesToSelect = DefaultIdField;
 
-            if (this.FieldsToSelect != null)
+            if (this.PropertiesToSelect != null)
             {
-                fieldsToSelect = this.FieldsToSelect.GetValue(state);
+                List<string> propertiesFound = this.PropertiesToSelect.GetValue(state);
+
+                if (propertiesFound != null && propertiesFound.Count > 0)
+                {
+                    propertiesToSelect = string.Join(",", propertiesFound);
+                }
             }
 
             string userId = this.UserId.GetValue(state);
 
             parameters.Add("UserId", userId);
             parameters.Add("MaxResults", maxCount);
-            parameters.Add("FieldsToSelect", fieldsToSelect);
+            parameters.Add("PropertiesToSelect", propertiesToSelect);
         }
 
         /// <inheritdoc />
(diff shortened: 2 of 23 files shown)


packages/Graph/Actions/User/GetDirectReports.cs
```diff
@@ -51,7 +51,11 @@ internal override async Task<IEnumerable<DirectoryObject>> CallGraphServiceWithR
             int maxCount = (int)parameters["MaxResults"];
 
             // Create the request first then apply the Top() after
-            IUserDirectReportsCollectionWithReferencesRequest request = client.Users[userId].DirectReports.Request();
+            // TODO: Make this SELECT() clause configurable for different column names
+            // EXPLAINER: The reason we are limiting the field is for two reasons:
+            //            1. Limit the size of the graph object payload needed to be transferred over the wire
+            //            2. Reduces the exposure of end user PII (Personally Identifiable Information) in our system for privacy reasons. It's generally good practice to use what you need.
+            IUserDirectReportsCollectionWithReferencesRequest request = client.Users[userId].DirectReports.Request().Select("id,displayName,mail,businessPhones,department,jobTitle,officeLocation");
```
peterinnesmsft, line 58: I don't think you should forcibly limit what fields are selected at this layer. This component should be designed for generic usage, and there may be valid reasons to require other fields. I think it's fine to have sensible defaults, but I would also ensure that it can overwritten.

I also wouldn't include the second comment - again, this is a generic component that others may use for their own purposes. There's no harm in interacting with PII as long as we understand what we are doing with it. We shouldn't lump in our own concerns with things like the Calendar, People and Enterprise Assistant template into this.

packages/Graph/Actions/User/GetManager.cs
```diff
@@ -37,7 +37,11 @@ public class GetManager : BaseMsGraphCustomAction<DirectoryObject>
         internal override async Task<DirectoryObject> CallGraphServiceWithResultAsync(IGraphServiceClient client, IReadOnlyDictionary<string, object> parameters, CancellationToken cancellationToken)
         {
             string userId = (string)parameters["UserId"];
-            DirectoryObject result = await client.Users[userId].Manager.Request().GetAsync(cancellationToken);
+            // TODO: Make this SELECT() clause configurable for different column names
+            // EXPLAINER: The reason we are limiting the field is for two reasons:
+            //            1. Limit the size of the graph object payload needed to be transferred over the wire
+            //            2. Reduces the exposure of end user PII (Personally Identifiable Information) in our system for privacy reasons. It's generally good practice to use what you need.
+            DirectoryObject result = await client.Users[userId].Manager.Request().Select("id,displayName,mail,businessPhones,department,jobTitle,officeLocation").GetAsync(cancellationToken);
```
peterinnesmsft, line 44: Same comments here. Also, if you're going to repeat the same fields that are selected, I would consider making this a bit more formal. Consider adding these strings as const values to some static class definition and then doing a string.Join on those fields rather than repeating it all over the place - it becomes a lot harder to keep in sync that way.

peterinnesmsft (changes requested):
Comments on the select fields being added to queries - please refactor to standardize so that strings aren't duplicated all over the place, and also ensure that they can be overwritten while providing sensible defaults.

packages/Graph/Actions/User/GetDirectReports.cs
thomas-chung: Makes sense. I just wanted to get unblocked on this first and then work on a second PR that includes a regenerated schema, etc. I guess I can make that change now.

packages/Graph/Schemas/Microsoft.Graph.User.GetDirectReports.schema
```diff
@@ -36,6 +36,14 @@
         "dialog.worksWithMe.maxResults"
       ]
     },
+    "fieldsToSelect": {
+      "$ref": "schema:#/definitions/stringExpression",
+      "title": "Fields To Select",
+      "description": "The fields of from the graph API to select. By default, if not provided only the ID will be returned for the object.",
+      "examples": [
+        "dialog.fieldsToSelect"
```
lauren-mills, line 44: Nit: might be nice to include examples here and also a link to where the field names can be found for people trying to use this for other use cases. Not critical though

lauren-mills (approved):

packages/Graph/Schemas/Microsoft.Graph.User.GetDirectReports.schema
thomas-chung: That's good idea! I've updated the description and examples to reflect that.

image

lauren-mills (approved):

packages/Graph/Actions/User/GetDirectReports.cs
```diff
@@ -41,6 +41,12 @@ public class GetDirectReports : BaseMsGraphCustomAction<IEnumerable<DirectoryObj
         [JsonProperty("MaxCount")]
         public IntExpression MaxCount { get; set; }
 
+        /// <summary>
+        /// Gets or sets the fields to select from the Graph API.
+        /// </summary>
+        [JsonProperty("FieldsToSelect")]
+        public StringExpression FieldsToSelect { get; set; }
```
peterinnesmsft, line 48: Should this not really be an ArrayExpression? And then do a join on the string values when passing through as a query parameter on the request itself?

packages/Graph/Actions/User/GetDirectReports.cs
```diff
@@ -80,10 +87,19 @@ protected override void PopulateParameters(DialogStateManager state, Dictionary<
                 maxCount = this.MaxCount.GetValue(state);
             }
 
+            // Select minimum of the "id" field from the object
+            string fieldsToSelect = DefaultIdField;
+
+            if (this.FieldsToSelect != null)
```
peterinnesmsft, line 93: This should then change to handling if the array is null or empty, then add the ID property name.

packages/Graph/Actions/BaseMsGraphCustomAction.cs
```diff
@@ -22,6 +22,11 @@ namespace Microsoft.Bot.Component.Graph.Actions
     /// </summary>
     public abstract class BaseMsGraphCustomAction : Dialog
     {
+        /// <summary>
+        /// Name of the id field that Graph uses to identify objects in the system.
+        /// </summary>
+        public const string DefaultIdField = "id";
```
peterinnesmsft, line 28: Rename to IdPropertyName. Alternatively, consider adding a PropertyNames static class and add `public const string Id = "id";` there, then reference that here.

packages/Graph/Actions/User/GetUserProfile.cs
```diff
@@ -50,9 +62,18 @@ protected override void PopulateParameters(DialogStateManager state, Dictionary<
                 throw new ArgumentNullException(nameof(this.UserId));
             }
 
+            // Select minimum of the "id" field from the object
+            string fieldsToSelect = DefaultIdField;
```
peterinnesmsft, line 66: This logic is being repeated in every single action. Is that really necessary? I don't know that we need to fix that this round, but this definitely feels like we have an opportunity to reduce code duplication.

packages/Graph/Actions/User/GetUserProfile.cs
```diff
@@ -30,14 +30,26 @@ public class GetUserProfile : BaseMsGraphCustomAction<User>
         [JsonProperty("UserId")]
         public StringExpression UserId { get; set; }
 
+        /// <summary>
+        /// Gets or sets the fields to select from the Graph API.
+        /// </summary>
+        [JsonProperty("FieldsToSelect")]
+        public StringExpression FieldsToSelect { get; set; }
```
peterinnesmsft, line 37: The Graph API documentation refers to these as properties - I would rename to SelectProperties or similar to match the existing nomenclature.

peterinnesmsft (changes requested):
A few more things to tweak, nothing major.

packages/Graph/Actions/User/GetUserProfile.cs
thomas-chung: Not necessarily every action - only ones that have to do with User objects. The duplication here is not too bad so I don't feel there is pressing need to refactor them.

packages/Graph/Actions/BaseMsGraphCustomAction.cs
thomas-chung: Do we really need a class for just one property right now?

thomas-chung: This is one of this thing where we can break it out into a different class if we see a need for it. And there isn't any adverse side effect if we don't do it.

packages/Graph/Actions/User/GetDirectReports.cs
thomas-chung: Sure.

peterinnesmsft (approved):

PR #854, opened 2021-04-06
Update calendar dev and test project

lauren-mills:
- Update calendar dev project and tests based on generator changes
- Rename to Calendar instead of calendarSkill


skills/declarative/Calendar/Calendar/Program.cs
```diff
@@ -4,7 +4,7 @@
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Hosting;
 
-namespace calendarskill
+namespace Calendar
```
thomas-chung, line 7: nit: just wondering if the namespace for our project should also include "Microsoft.Bot.*"

thomas-chung (approved):

PR #856, opened 2021-04-06
Update people dev project and tests based on generator

lauren-mills:
Updating People dev project to match the generator and renaming from whoSkill to People


thomas-chung (approved):
Before you push this should look at my renames as well (I've renamed dialogs to more sane "Get X" instead of the "Who Is X")

PR #882, opened 2021-04-12
Add Teams SSOTokenExchange

EricDahlvang:
### Purpose
If the activity name is signin/tokenExchange, this middleware will attempt to exchange the token, and deduplicate the incoming call, ensuring only one exchange request is processed.

### Tests
will add these

#### General


packages/Teams/Middleware/SSOTokenExchangeMiddleware.cs
```diff
@@ -0,0 +1,210 @@
...
+        private readonly string _oAuthConnectionName;
+
+        /// <summary>
+        /// Creates an instance of <see cref="SSOTokenExchangeMiddleware" />.
+        /// </summary>
+        /// <param name="configuration"></param>
+        /// <param name="storage"></param>
+        public SSOTokenExchangeMiddleware(IConfiguration configuration, IStorage storage)
+        {
+            _oAuthConnectionName = configuration[ConnectionNameProperty];
```
peterinnesmsft, line 47: Part of the reason for the component settings design is that ideally this stuff can come from the settings for the component. If we take IConfiguration, it's going to always expect ConnectionName to be at the root of the IConfiguration object, which is less great to me because it monopolizes the shape of the user's settings.

packages/Teams/Middleware/SSOTokenExchangeMiddleware.cs
```diff
@@ -0,0 +1,210 @@
...
+    /// 
+    /// Only one of these token exchange requests should be processed by the bot.
+    /// The others return <see cref="System.Net.HttpStatusCode.PreconditionFailed"/>.
+    /// For a distributed bot in production, this requires a distributed storage
+    /// ensuring only one token exchange is processed. This middleware supports
+    /// CosmosDb storage found in Microsoft.Bot.Builder.Azure, or MemoryStorage for
+    /// local development. IStorage's ETag implementation for token exchange activity
+    /// deduplication.
+    /// </remarks>
+    public class SSOTokenExchangeMiddleware : IMiddleware
```
peterinnesmsft, line 34: Should we make this internal? I'm fine with registering the middleware by hooking this up as a component, but not sure if there are unintended implications if we expose this outside of the assembly otherwise.

packages/Teams/Middleware/SSOTokenExchangeMiddleware.cs
```diff
@@ -0,0 +1,210 @@
...
+            }
+            else
+            {
+                // NOTE: this could be used in conjunction with an OAuthInput dialog which checks the 
+                // TurnState for the responses, rather than re-creating them which results in two exchange request calls.
+
+                //// Store response in TurnState, so the TokenExchangeOAuthPrompt can use it, and not have to do the exchange again.
+                //turnContext.TurnState[nameof(TokenExchangeInvokeRequest)] = tokenExchangeRequest;
+                //turnContext.TurnState[nameof(TokenResponse)] = tokenExchangeResponse;
+            }
```
peterinnesmsft, line 180: It doesn't look like this is adding anything. Remove?

packages/Teams/Middleware/SSOTokenExchangeMiddleware.cs
```diff
@@ -0,0 +1,210 @@
...
+            try
+            {
+                // Writing the IStoreItem with ETag of unique id will succeed only once
+                await _storage.WriteAsync(storeItems, cancellationToken).ConfigureAwait(false);
+            }
+            catch (Exception ex)
+
+                // Memory storage throws a generic exception with a Message of 'Etag conflict. [other error info]'
+                // CosmosDbPartitionedStorage throws: ex.Message.Contains("pre-condition is not met")
+                when (ex.Message.StartsWith("Etag conflict") || ex.Message.Contains("pre-condition is not met"))
```
peterinnesmsft, line 91: This seems a little unreliable. Can we not check for a particular exception type and some kind of error code instead?

packages/Teams/Middleware/SSOTokenExchangeMiddleware.cs
```diff
@@ -0,0 +1,210 @@
...
+        {
+            TokenResponse tokenExchangeResponse = null;
+            var tokenExchangeRequest = ((JObject)turnContext.Activity.Value)?.ToObject<TokenExchangeInvokeRequest>();
+
+            try
+            {
+                var userTokenClient = turnContext.TurnState.Get<UserTokenClient>();
+                if (userTokenClient != null)
+                {
+                    tokenExchangeResponse = await userTokenClient.ExchangeTokenAsync(turnContext.Activity.From.Id,
```
peterinnesmsft, line 121: This indentation looks way off.

packages/Teams/Middleware/SSOTokenExchangeMiddleware.cs
```diff
@@ -0,0 +1,210 @@
...
+                                                                    _oAuthConnectionName,
+                                                                    turnContext.Activity.ChannelId,
+                                                                    new TokenExchangeRequest { Token = tokenExchangeRequest.Token },
+                                                                    cancellationToken).ConfigureAwait(false);
+                }
+                else if (turnContext.Adapter is IExtendedUserTokenProvider adapter)
+                {
+                    tokenExchangeResponse = await adapter.ExchangeTokenAsync(
```
peterinnesmsft, line 129: Same here.

packages/Teams/Middleware/SSOTokenExchangeMiddleware.cs
```diff
@@ -0,0 +1,210 @@
...
+#pragma warning disable CA1031 // Do not catch general exception types (ignoring, see comment below)
+            catch
+#pragma warning restore CA1031 // Do not catch general exception types
+            {
+                // Ignore Exceptions
+                // If token exchange failed for any reason, tokenExchangeResponse above stays null,
+                // and hence we send back a failure invoke response to the caller.
+            }
+
+            if (tokenExchangeResponse == null || string.IsNullOrEmpty(tokenExchangeResponse.Token))
```
peterinnesmsft, line 150: `if (string.IsNullOrEmpty(tokenExchangeResponse?.Token))`

packages/Teams/Middleware/SSOTokenExchangeMiddleware.cs
```diff
@@ -0,0 +1,210 @@
...
+        {
+            public string ETag { get; set; }
+
+            public static string GetStorageKey(ITurnContext turnContext)
+            {
+                var activity = turnContext.Activity;
+                var channelId = activity.ChannelId ?? throw new InvalidOperationException("invalid activity-missing channelId");
+                var conversationId = activity.Conversation?.Id ?? throw new InvalidOperationException("invalid activity-missing Conversation.Id");
+
+                if (activity.Type != ActivityTypes.Invoke || activity.Name != SignInConstants.TokenExchangeOperationName)
```
peterinnesmsft, line 195: Any reason for not doing case-insensitive checks?

peterinnesmsft (changes requested):
Some questions on the wiring up of the component as well as implementation of the middleware.

packages/Teams/TeamsBotComponent.cs
```diff
@@ -15,6 +16,22 @@ public class TeamsBotComponent : BotComponent
...
+
+            if (configuration == null)
+            {
+                throw new ArgumentNullException(nameof(configuration));
+            }
+
+            var ettings = configuration.GetSection(ComponentSettings.SettingsKey).Get<ComponentSettings>() ?? new ComponentSettings();
+            if (ettings.SSOMiddleware)
+            {
+                services.AddSingleton<IMiddleware, SSOTokenExchangeMiddleware>();
```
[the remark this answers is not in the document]
EricDahlvang: This will be picked up by the adapter.

SSOTokenExchangeMiddleware is a short-circuit middleware, so will affect processing of other middlewares after it in the stack.  This is a known issue with middleware.

clearab:
@EricDahlvang this can be closed now, right?

EricDahlvang:
@clearab  no, we still need the adaptive pieces this provides (registration of the middleware)

packages/Teams/Middleware/SSOTokenExchangeMiddleware.cs
EricDahlvang: I don't like this either.

peterinnesmsft (approved):

PR #916, opened 2021-04-16
Adaptive generator: add reference to luis + qna for r14 legacy story

carlosscastro:
Currently the runtime has a set of built-in components that cannot be removed and are statically registered. Many of these make sense even in the super long term such as AdaptiveDialogs or BotBuilder. However, other of the core components that we've always registered are Luis and QnA. With the new component model, in R14 we would want to remove Luis and QnA and have them be only registered as components by the generators that need them. This will decouple the runtime from LUIS + QnA and also allow users to not have dependencies they won't want.
 
This change prepares the terrain for that to reduce breaking changes and need for migrations in R13 generators when upgrading to r14.


tracyboehrer (approved):

tracyboehrer:
@peterinnesmsft @carlosscastro We'll need to cherry pick this to the 4.13 too.

PR #922, opened 2021-04-16
Update calendar OAuth implementation and clear data from traceActivities

lauren-mills:
### Changes
- Update OAuth throughout to use turn.token and refresh before each graph call
- Set value of trace activities to null to prevent any extra data sent over the wire
- Cleanup leftover dogfood logic (dialogComplete & collectFeedback)
- Update custom action schemas to use turn.token.token as example to promote best practices

### DIFF
--- a/skills/declarative/Calendar/Calendar/dialogs/CreateEventDialog/CreateEventDialog.dialog
+++ b/skills/declarative/Calendar/Calendar/dialogs/CreateEventDialog/CreateEventDialog.dialog
@@ -21,7 +21,8 @@
           "$designer": {
             "id": "mBFKDC"
           },
-          "name": "CreateEventDialog.BeginDialog"
+          "name": "CreateEventDialog.BeginDialog",
+          "value": "=null"
         },
         {
           "$kind": "Microsoft.SetProperty",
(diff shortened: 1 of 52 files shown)


skills/declarative/Calendar/Calendar/dialogs/CreateEventDialog/CreateEventDialog.dialog
```diff
@@ -332,15 +335,24 @@
           "$designer": {
             "id": "2yLHBA"
           },
-          "name": "CreateEventDialog.OnEvent_FindMeetingTimes"
+          "name": "CreateEventDialog.OnEvent_FindMeetingTimes",
+          "value": "=null"
```
thomas-chung, line 339: Is this actually needed? I thought it would just default to null anyway

thomas-chung (approved):

skills/declarative/Calendar/Calendar/dialogs/CreateEventDialog/CreateEventDialog.dialog
lauren-mills: No, when its not provided, it sends the full state read out

PR #934, opened 2021-04-19
Tweaked trigger names displayed in composer

### LINKED ISSUE #926: Adaptive Cards: form titles for triggers need cleanup
The titles displayed for Adaptive Card Triggers need some cleanup:

Stevenic:
Resolves #926 

Updated the .uischema for triggers to better align with Composer conventions naming wise.


clearab (approved):
Looks good to me, but may want to consider dropping "activity" from those labels so it is just "Action.Execute (invoke)" and "Action.Submit (message)" - message is a bit awkward, but in context I think it would be ok.

peterinnesmsft (approved):

PR #941, opened 2021-04-20
feat: .NET Functions Static File support

peterinnesmsft:
### Purpose
For skills-based scenarios, we need to support serving static files from the .NET Functions-based bots, for content such as manifest, .lu and .qna files. This PR adds an HTTP trigger out of the box to support this scenario.

### Changes
- Update host.json to change the http route prefix to "" from the default of "api".
- Updating MessagesTrigger and SkillsTrigger to add "api" to routes.
- Adding StaticFilesTrigger to handle returning of supported static files.
- Update scaffolded .csproj to include all files within 'www' directory as part of build drop.

### Tests
Manually tested via local Functions app, as well as published Functions app on Azure.


generators/generator-bot-adaptive/generators/app/templates/dotnet/functions/Triggers/StaticFilesTrigger.cs
```diff
@@ -0,0 +1,66 @@
...
+            if (!TryGetFilePath(context, path, out string filePath) ||
+                !_provider.TryGetContentType(filePath, out string contentType))
+            {
+                return new NotFoundResult();
+            }
+
+            return new FileStreamResult(new FileStream(filePath, FileMode.Open), contentType);
+        }
+
+        bool TryGetFilePath(ExecutionContext context, string relativePath, out string result)
```
thomas-chung, line 41: nit: private

generators/generator-bot-adaptive/generators/app/templates/dotnet/functions/Triggers/StaticFilesTrigger.cs
```diff
@@ -0,0 +1,66 @@
...
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.StaticFiles;
+using Microsoft.Azure.WebJobs;
+using Microsoft.Azure.WebJobs.Extensions.Http;
+
+namespace <%= botName %>.Triggers
+{
+    public class StaticFilesTrigger
+    {
+        private const string StaticFilesDirectory = "www";
```
thomas-chung, line 13: Should this be "wwwroot" instead so it's inline with the static file version counterpart in the web app version?

generators/generator-bot-adaptive/generators/app/templates/dotnet/functions/Triggers/StaticFilesTrigger.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        private readonly FileExtensionContentTypeProvider _provider;
+
+        public StaticFilesTrigger()
+        {
+            var provider = new FileExtensionContentTypeProvider();
+            provider.Mappings[".lu"] = "application/vnd.microsoft.lu";
+            provider.Mappings[".qna"] = "application/vnd.microsoft.qna";
```
thomas-chung, line 21: What about manifest files extensions? Do we need to provide that as well?

joshgummersall (approved):
Looks functionally consistent with JS.

generators/generator-bot-adaptive/generators/app/templates/dotnet/functions/Triggers/StaticFilesTrigger.cs
thomas-chung: I guess the flip side is on Azure Function the web app seems to be on wwwroot already so it's like wwwroot/wwwroot.

thomas-chung (approved):

tracyboehrer (approved):

generators/generator-bot-adaptive/generators/app/templates/dotnet/functions/Triggers/StaticFilesTrigger.cs
peterinnesmsft: FileExtensionContentTypeProvider contains a default mapping for many different file extensions to corresponding content types. As the schema files are just .json, these are already covered by the default mapping. The way this code is written just supplements that by adding the additional ones needed for our scenarios.

generators/generator-bot-adaptive/generators/app/templates/dotnet/functions/Triggers/StaticFilesTrigger.cs
peterinnesmsft: Yeah, it's a little weird for sure. Looking at other implementations in places, this seems to be the way it's been handled. At the end of the day, customers can easily change it to something a little more fitting should they wish.

PR #942, opened 2021-04-20
Adding Telephony preview package, this includes Call Transfer Activit…

vanshu86:
…y currently and it will be a place for all telephony extensions for composer

### Purpose
Adding Telephony preview package, this includes Call Transfer Activity currently and it will be a place for all telephony extentions for composer

### Changes
Enabling Telephony Features in Composer via component package

### Tests
Manual test of creating bot with telephony package and verified behavior

### Feature Plan
More features to be  added as we go along

#### General


packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
```diff
@@ -0,0 +1,53 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using AdaptiveExpressions.Properties;
```
carlosscastro, line 4: Sorting: system first, then alphabetical

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
```diff
@@ -0,0 +1,53 @@
...
+        public StringExpression TargetPhoneNumber { get; set; }
+
+        public async override Task<DialogTurnResult> BeginDialogAsync(DialogContext dc, object options = null, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            var targetPhoneNumber = TargetPhoneNumber.GetValue(dc.State);
+
+            await dc.Context.SendActivityAsync($"Transferring to \"{targetPhoneNumber}\"...");
+
+            // Create handoff event, passing the phone number to transfer to as context.
+            var poContext = new { TargetPhoneNumber = targetPhoneNumber };
```
carlosscastro, line 37: what's 'po'? In general we try to use full words except in obvious acronyms such as 'http'

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
```diff
@@ -0,0 +1,53 @@
...
+            var poHandoffEvent = EventFactory.CreateHandoffInitiation(dc.Context, poContext);
+
+            try
+            {
+                await dc.Context.SendActivityAsync(poHandoffEvent, cancellationToken);
+                await dc.Context.SendActivityAsync($"Call transfer initiation succeeded");
```
carlosscastro, line 43: This is not a good idea. Text should not be hardcoded (what about other languages? Also, would composer users want to control what activities are sent or other actions that are performed? In that case, this dialog would be drastically different. We can talk in the design call :)

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
```diff
@@ -0,0 +1,53 @@
...
+            }
+            catch
+            {
+                await dc.Context.SendActivityAsync($"Call transfer failed");
```
carlosscastro, line 47: No hardcoded text pls. Also we should be catching the exception and making sure it goes to app insights. What exception types could we get from call transfer?

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
```diff
@@ -0,0 +1,53 @@
...
+            }
+
+            return await dc.EndDialogAsync(result: 0, cancellationToken: cancellationToken);
```
carlosscastro, line 50: why is result `0`? What does that mean?

packages/Microsoft.Bot.Components.Telephony/Microsoft.Bot.Components.Telephony.csproj
```diff
@@ -0,0 +1,50 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Version Condition=" '$(PreviewPackageVersion)' == '' ">$(LocalPackageVersion)</Version>
```
carlosscastro, line 4: These build variables exist inthe SDK but not in this repo. Check out the graph csproj to be consistent in how versioning works here.

packages/Microsoft.Bot.Components.Telephony/Microsoft.Bot.Components.Telephony.csproj
```diff
@@ -0,0 +1,50 @@
...
+    <Version Condition=" '$(PreviewPackageVersion)' != '' ">$(PreviewPackageVersion)</Version>
+    <PackageVersion Condition=" '$(PreviewPackageVersion)' == '' ">$(LocalPackageVersion)</PackageVersion>
+    <PackageVersion Condition=" '$(PreviewPackageVersion)' != '' ">$(PreviewPackageVersion)</PackageVersion>
+    <Configurations>Debug;Release</Configurations>
+    <DocumentationFile>bin\$(Configuration)\netstandard2.0\Microsoft.Bot.Builder.Dialogs.Adaptive.xml</DocumentationFile>
```
carlosscastro, line 9: Update to right name

packages/Microsoft.Bot.Components.Telephony/Microsoft.Bot.Components.Telephony.csproj
```diff
@@ -0,0 +1,50 @@
...
+  <ItemGroup>
+    <Content Include="**/*.dialog" />
+    <Content Include="**/*.lg" />
+    <Content Include="**/*.lu" />
+    <Content Include="**/*.schema" />
+    <Content Include="**/*.uischema" />
+    <Content Include="**/*.qna" />
+  </ItemGroup>
+
+  <ItemGroup>
```
carlosscastro, line 42: remove this group

packages/Microsoft.Bot.Components.Telephony/Microsoft.Bot.Components.Telephony.csproj
```diff
@@ -0,0 +1,50 @@
...
+    <None Remove="Schemas\CallTransferDialog.schema" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Bot.Builder.Dialogs.Adaptive" Version="4.13.0-rc4" />
```
carlosscastro, line 47:
```suggestion
    <PackageReference Include="Microsoft.Bot.Builder.Dialogs.Adaptive" Version="4.13.0" />
```

packages/Microsoft.Bot.Components.Telephony/Schemas/CallTransferDialog.schema
```diff
@@ -0,0 +1,18 @@
...
+  "$role": "implements(Microsoft.IDialog)",
+  "title": "Call Transfer",
+  "description": "Phone number to transfer the call to (in E.164 format such as +1425123456)",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "targetPhoneNumber": {
+      "$ref": "schema:#/definitions/stringExpression",
+      "title": "Target Phone Number",
+      "description": "Phone number to transfer the call to",
```
carlosscastro, line 12: Description ideally should end with a period

```suggestion
      "description": "Phone number to transfer the call to.",
```

packages/Microsoft.Bot.Components.Telephony/Schemas/CallTransferDialog.uischema
```diff
@@ -0,0 +1,26 @@
...
+      "targetPhoneNumber": {
+        "intellisenseScopes": [
+          "variable-scopes"
+        ]
+      }
+    }
+  },
+  "menu": {
+    "label": "Transfer a call",
+    "submenu": [ "Channels", "Telephony" ]
```
carlosscastro, line 20: We don't use the word channels in composer, we try to use Connections. I can run this by @sgellock / @benbrown / design on the wording here.

packages/Microsoft.Bot.Components.Telephony/TelephonyBotComponent.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.Bot.Components.Telephony.Actions;
+using Microsoft.Bot.Builder.Dialogs.Declarative;
+using Microsoft.Extensions.Configuration;
```
carlosscastro, line 6: alphabetical sort

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
```diff
@@ -0,0 +1,53 @@
...
+        [JsonConstructor]
+        public CallTransferDialog([CallerFilePath] string sourceFilePath = "", [CallerLineNumber] int sourceLineNumber = 0)
+            : base()
+        {
+            // enable instances of this command as debug break point
+            this.RegisterSourceLocation(sourceFilePath, sourceLineNumber);
+        }
+
+        [JsonProperty("$kind")]
+        public const string Kind = "CallTransferDialog";
```
carlosscastro, line 25: Proposed name: 

```suggestion
        public const string Kind = "Microsoft.Telephony.CallTransfer";
```

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        [JsonProperty("targetPhoneNumber")]
+        public StringExpression TargetPhoneNumber { get; set; }
```
carlosscastro, line 28: Naming: Should we consider just `PhoneNumber`?

packages/Microsoft.Bot.Components.Telephony/Schemas/CallTransferDialog.schema
```diff
@@ -0,0 +1,18 @@
+﻿{
+  "$schema": "https://schemas.botframework.com/schemas/component/v1.0/component.schema",
+  "$role": "implements(Microsoft.IDialog)",
+  "title": "Call Transfer",
+  "description": "Phone number to transfer the call to (in E.164 format such as +1425123456)",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "targetPhoneNumber": {
```
carlosscastro, line 9: phone number should be required

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        public async override Task<DialogTurnResult> BeginDialogAsync(DialogContext dc, object options = null, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            var targetPhoneNumber = TargetPhoneNumber.GetValue(dc.State);
```
carlosscastro, line 32: Null check?

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
```diff
@@ -0,0 +1,53 @@
...
+
+            await dc.Context.SendActivityAsync($"Transferring to \"{targetPhoneNumber}\"...");
```
carlosscastro, line 34: Cannot hardcode strings

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
```diff
@@ -0,0 +1,53 @@
...
+
+            // Create handoff event, passing the phone number to transfer to as context.
+            var poContext = new { TargetPhoneNumber = targetPhoneNumber };
+            var poHandoffEvent = EventFactory.CreateHandoffInitiation(dc.Context, poContext);
+
+            try
+            {
+                await dc.Context.SendActivityAsync(poHandoffEvent, cancellationToken);
```
carlosscastro, line 42:
```suggestion
                await dc.Context.SendActivityAsync(poHandoffEvent, cancellationToken).ConfigureAwait(false);
```

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
carlosscastro, line 50:
```suggestion
            return await dc.EndDialogAsync(result: 0, cancellationToken: cancellationToken).ConfigureAwait(false);
```

carlosscastro:
Thanks Vanshika! Great progress here. Some general details and also we have to discuss the action experience design before moving forward. Shouldn't be too bad to get this into a mergeable state once we have the experience on point.

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
vanshu86: I had changed my $kind from CallTransferDialog to Microsoft.Bot.Components.CallTransferDialog and that was causing the issue --  I hear the message that my dialog is not registered with factory. With CallTransferDialog it works

packages/Microsoft.Bot.Components.Telephony/TelephonyBotComponent.cs
vanshu86: done

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
vanshu86: done

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
vanshu86: what's the alternative?

packages/Microsoft.Bot.Components.Telephony/Microsoft.Bot.Components.Telephony.csproj
vanshu86: which group?

packages/Microsoft.Bot.Components.Telephony/Microsoft.Bot.Components.Telephony.csproj
vanshu86: are these autogenerated xmls? sorry whats the right name?

packages/Microsoft.Bot.Components.Telephony/Microsoft.Bot.Components.Telephony.csproj
vanshu86: ok I took it from the Teams csproj will now look at the graph one.

packages/Microsoft.Bot.Components.Telephony/Microsoft.Bot.Components.Telephony.csproj
vanshu86: so graph project doesnt have this so removed

packages/Microsoft.Bot.Components.Telephony/Microsoft.Bot.Components.Telephony.csproj
vanshu86: removed

packages/Microsoft.Bot.Components.Telephony/Schemas/CallTransferDialog.uischema
vanshu86: We want to make it channels as these actions are Channel specific - discussed this with Steve

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
vanshu86: we need to notify the user somehow that it has failed, whats the best way we can do that? 
I have added this.TelemetryClient.TrackException(ex); for insights, unless there is another recommended way

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
vanshu86: sounds good

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
vanshu86: sure I think that should be fine

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
vanshu86: removed po

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
vanshu86: made this null, I think this is because it doesnt need to pass back any parameter

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
carlosscastro, line 4: System.* comes first, and then all others. All the System.* should be alphabetical amont them and all the non-system after those. All the sDK files have this ordering if you want to see examples

carlosscastro: Example

```csharp
using System.Runtime.CompilerServices;
using System.Threading;
using System.Threading.Tasks;
using AdaptiveExpressions;
using Microsoft.Bot.Builder;
using Microsoft.Bot.Builder.Dialogs;
using Newtonsoft.Json;
```

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
```diff
@@ -0,0 +1,53 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using AdaptiveExpressions.Properties;
+using Microsoft.Bot.Builder;
+using Microsoft.Bot.Builder.Dialogs;
+using Newtonsoft.Json;
+using System.Runtime.CompilerServices;
+using System.Threading;
```
carlosscastro, line 9: Do you really need this one?

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
carlosscastro: Yes, because you also need to update your schema file name and the CallTransferDialog.Kind property to the same thing.

carlosscastro: Lets get the right name, and then we can help you make it work. But it must start with Microsoft.*

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
carlosscastro: We can discuss this one with the team, leaving unresolved so we discuss before merging

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
carlosscastro: Check this out: https://github.com/microsoft/botbuilder-dotnet/blob/main/libraries/Microsoft.Bot.Builder.Dialogs.Adaptive/Input/InputDialog.cs#L105

Allows authors to specify the activity that gets sent, including whether its a card or text, whether it comes from the language generation system (and thus getitng free localization), etc

carlosscastro: This should be optional also, since some may not want to send an activity.

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
carlosscastro: I still see it, unresolving so we track it.

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
carlosscastro: Check https://github.com/microsoft/botbuilder-dotnet/blob/main/libraries/Microsoft.Bot.Builder.Dialogs.Adaptive/Input/InputDialog.cs#L105

However, I would like to see a scenario design for this. Do we really want to send an activity here?

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
carlosscastro: I'm not sure that the right thing is to notify the user. Would it be better to notify the user or to signal the failure so that the bot author can take action based on the failure? Seems like there are learnings to take from the discussion on STatRecording here.

ryanlengel (changes requested):
Can we rename the project directory to packages/Telephony? At least for now that's how we do it with the other packages since they could be in multiple languages/different namespaces.

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
vanshu86: yes

vanshu86: yes

packages/Microsoft.Bot.Components.Telephony/Actions/CallTransferDialog.cs
vanshu86: This has been updated already resolving now

carlosscastro (approved):

carlosscastro:
Discussed offline with @ryanlengel  before merging :) This is a feature branch and we addressed his feedback

PR #965, opened 2021-04-22
[EA] Update authentication docs

### LINKED ISSUE #961: Update docs for enterprise assistant provisioning
image

lauren-mills:
Fixes #961 

Updates the authentication setup instructions for EA, Calendar, and People

### DIFF
--- a/generators/generator-bot-calendar/generators/app/templates/README.md
+++ b/generators/generator-bot-calendar/generators/app/templates/README.md
@@ -40,7 +40,7 @@ You must configure an authentication connection on your Azure Bot Registration i
 
 9. In the Authentication tab, click **Add a platform**
     1. Select **Web**
-    2. Set the URL to https://token.botframework.com/.auth/web/redirect.
+    2. Set the URL to https://token.botframework.com/.auth/web/redirect
 10. In Bot Framework Composer, open your **Project Settings** and toggle the **Advanced Settings View**
 1. Set the following property to the value from Step 3:
     ```
--- a/generators/generator-bot-calendar/generators/app/templates/README.md
+++ b/generators/generator-bot-calendar/generators/app/templates/README.md
@@ -10,6 +10,13 @@ In order to test this bot locally, you will need the following services provisio
 - Azure Bot Registration
 - Language Understanding (LUIS)
 
+#### Configure Microsoft App Password
+1. Open your Azure Bot Channels Regisration in the Azure Portal
+2. In the **Configuration** tab, click **Manage** next to your Microsoft App ID
+3. In the Certificates & secrets tab, click **New client secret**
+4. Assign a name and an expiration period, then click **Add**
+5. Copy the secret value and save for later use along with your Microsoft App ID
+
 ## Configure Authentication
 You must configure an authentication connection on your Azure Bot Registration in order to log in and access Microsoft Graph resources. You can configure these settings either through the Azure Portal or via the Azure CLI.
 
--- a/generators/generator-bot-calendar/generators/app/templates/README.md
+++ b/generators/generator-bot-calendar/generators/app/templates/README.md
@@ -3,9 +3,11 @@ This folder contains a Bot Project created with Bot Framework Composer.
 The full documentation for Composer lives here:
 https://github.com/microsoft/botframework-composer
 
-To test this bot locally, follow these instructions:
+To test this bot locally, you will need to complete the following steps:
+1. Provision your Azure resources for local development
+2. Configure your authentication settings
 
-## Provision Azure Resources to Host Bot
+## 1. Provision Azure Resources
 In order to test this bot locally, you will need the following services provisioned in Azure:
 - Azure Bot Registration
 - Language Understanding (LUIS)
@@ -17,7 +19,7 @@ In order to test this bot locally, you will need the following services provisio
 4. Assign a name and an expiration period, then click **Add**
 5. Copy the secret value and save for later use along with your Microsoft App ID
 
-## Configure Authentication
+## 2. Configure Authentication
 You must configure an authentication connection on your Azure Bot Registration in order to log in and access Microsoft Graph resources. You can configure these settings either through the Azure Portal or via the Azure CLI.
 
 ### Option 1: Using the Azure Portal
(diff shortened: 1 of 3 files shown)


generators/generator-bot-calendar/generators/app/templates/README.md
```diff
@@ -10,25 +10,46 @@ In order to test this bot locally, you will need the following services provisio
 - Azure Bot Registration
 - Language Understanding (LUIS)
 
+#### Configure Microsoft App Password
```
thomas-chung, line 13: What do you think of giving people a high level topic/summary of all the steps needed at the beginning of this doc. This is just to prep people on basically at a high level what they should expect as far as setup is concerned. 

This way I think it can also prevent people from feeling they are stuck in "manual hell" where they just keep following steps with no end in sight when they are half way through?

generators/generator-bot-calendar/generators/app/templates/README.md
```diff
@@ -10,25 +10,46 @@ In order to test this bot locally, you will need the following services provisio
...
+7. In the API permissions tab, click **Add a permission**
+8. Click **Microsoft Graph** and add the following scopes: 
+    * Calendars.ReadWrite
+    * Contacts.Read
+    * People.Read
+    * User.ReadBasic.All
+
+9. In the Authentication tab, click **Add a platform**
+    1. Select **Web**
+    2. Set the URL to https://token.botframework.com/.auth/web/redirect.
```
thomas-chung, line 43: nit: you might want to not have a period at the end of this in case people make mistakes copy and pasting.

generators/generator-bot-calendar/generators/app/templates/README.md
```diff
@@ -10,25 +10,46 @@ In order to test this bot locally, you will need the following services provisio
...
+10. In Bot Framework Composer, open your **Project Settings** and toggle the **Advanced Settings View**
+1. Set the following property to the value from Step 3:
+    ```
+    {
+      "oauthConnectionName": "Outlook",
```
thomas-chung, line 48: You may want to specify where in the json to add it to, cause the object is nested you might want to specifically direct them to it. (Maybe include a couple other lines from the same level in here too as a hint of some sort, like "MicrosoftAppId")

thomas-chung: We may also want to tell them if they don't have the setting configured correctly what error they should expect. (see my PR for the emit trace for example)

thomas-chung: We should really start thinking about frequently encountered problems and start documenting them here as they come in.

generators/generator-bot-enterprise-assistant/generators/app/templates/README.md
```diff
@@ -43,13 +127,13 @@ You must configure an authentication connection on your **skill bot** Azure Bot
...
+    az rest --method patch --url https://graph.microsoft.com/v1.0/applications/<objectId> --body "{ 'requiredResourceAccess': [{'resourceAppId': '00000003-0000-0000-c000-000000000000','resourceAccess': [{ 'id': 'b89f9189-71a5-4e70-b041-9887f0bc7e4a', 'type': 'Scope' }, { 'id': 'b340eb25-3456-403f-be2f-af7a0d370277',	'type': 'Scope'	}, { 'id': 'a154be20-db9c-4678-8ab7-66f6cc099a59',	'type': 'Scope'	}, { 'id': '06da0dbc-49e2-44d2-8312-53f166ab848a', 'type': 'Scope' }, { 'id': 'ff74d97f-43af-4b68-9f2a-b77ee6968c5d', 'type': 'Scope'	}, { 'id': 'ba47897c-39ec-4d83-8086-ee8256fa737d', 'type': 'Scope' } ]}	]}"
```
thomas-chung, line 130: What are these IDs? I assume these IDs are simply the identifier for the scopes. We may want to map them to the actual textual name.

generators/generator-bot-enterprise-assistant/generators/app/templates/README.md
```diff
@@ -72,17 +156,16 @@ After provisioning your Azure resources and configuring your authentication conn
     - In **Advanced Settings View (json)**, set the app IDs of the skills that will be allowed to communicate with your root bot. For testing you can set this to "*".
```
thomas-chung, line 156: It would be nice if we can also tell them to not use "*" when deploying because that opens the bot up for anyone to call and call it out as a security risk. Developers are a lazy bunch (me included), once something works there is never really a reason to change it until something breaks. :)

thomas-chung (changes requested):

generators/generator-bot-enterprise-assistant/generators/app/templates/README.md
lauren-mills: These are the IDs. Since I don't have a script file for them to run, I didn't want to add any overhead as far as an additional mapping method goes. Let me know if you have ideas for how to communicate that here

thomas-chung: I think having a simple table or bulleted list underneath works. 

For example:

> Below are mapping between each IDs mentioned above and their Permission Name in Microsoft Graph.
> - People.Read.All = b89f9189-71a5-4e70-b041-9887f0bc7e4a
> - User.ReadBasic.All = b340eb25-3456-403f-be2f-af7a0d370277
> - ...

peterinnesmsft (approved):

generators/generator-bot-calendar/generators/app/templates/README.md
lauren-mills: The setting already exists in the templates, its just blank. I can't consistently show where it will show in the appsettings file since it gets added by the generator

generators/generator-bot-enterprise-assistant/generators/app/templates/README.md
lauren-mills: Lets hold off on this addition for now. I'll add it next week

generators/generator-bot-enterprise-assistant/generators/app/templates/README.md
lauren-mills: Removed the * text

generators/generator-bot-calendar/generators/app/templates/README.md
lauren-mills: Removed

generators/generator-bot-calendar/generators/app/templates/README.md
lauren-mills: Added

thomas-chung (approved):

PR #974, opened 2021-04-23
Teams: Added CommandId property to applicable messaging extension triggers

### LINKED ISSUE #925: Teams: Message Extension triggers should support an optional "commandId" property

tracyboehrer:
Fixes #925

### DIFF
--- a/packages/Teams/TriggerConditions/OnTeamsMEBotMessagePreviewEdit.cs
+++ b/packages/Teams/TriggerConditions/OnTeamsMEBotMessagePreviewEdit.cs
@@ -17,18 +17,18 @@ namespace Microsoft.Bot.Components.Teams.Conditions
     /// </summary>
     public class OnTeamsMEBotMessagePreviewEdit : OnInvokeActivity
     {
-        [JsonProperty("$kind")]
-        public new const string Kind = "Teams.OnMEBotMessagePreviewEdit";
-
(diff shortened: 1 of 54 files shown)


tracyboehrer:
@Stevenic I have been unable to determine if the messaging extensions for SelectItem, CardButtonClicked, and ConfigSetting also have this CommandId.  In Dotnet, the data is just passed as an object.  Still researching.

packages/Teams/TriggerConditions/OnTeamsMEBotMessagePreviewEdit.cs
```diff
@@ -20,17 +20,32 @@ public class OnTeamsMEBotMessagePreviewEdit : OnInvokeActivity
         [JsonProperty("$kind")]
         public new const string Kind = "Teams.OnMEBotMessagePreviewEdit";
 
+        [JsonProperty("commandId")]
```
peterinnesmsft, line 23: Properties after constructor please. Apply throughout.

peterinnesmsft (changes requested):
Overall looks fine - just place properties after constructors.

peterinnesmsft (approved):

PR #978, opened 2021-04-23
Update templates to latest

lauren-mills:
Update EA, Calendar, and People templates with latest


generators/generator-bot-calendar/generators/app/templates/dialogs/CreateEventDialog/CreateEventDialog.dialog
```diff
@@ -149,7 +150,8 @@
           "$designer": {
             "id": "DHvA61"
           },
-          "name": "CreateEventDialog.OnEvent_SetTitle"
+          "name": "CreateEventDialog.OnEvent_SetTitle",
+          "value": "=null"
```
thomas-chung, line 154: If this forces every skill developers to the add a "=null" to their bot, then it would be tragic. :(

thomas-chung: Note to self - bring this up with the SDK team. Not as a P0 but probably as a P1 issue to fix.

thomas-chung (approved):

Issue #973, opened 2021-04-23
Enterprise assistant: Provide more actionable message for the "Operation returned an invalid status code Unauthorized" error


elaine-chang:
## Version
![image](https://user-images.githubusercontent.com/25210203/115826131-482e3180-a3bf-11eb-8f3a-9a692101ee8c.png)

## Describe the bug
User will very easily hit the Unauthorized error just following the current creation flow to create EA and start locally. Recommend to provide a link to configure authentication to help unblock user in this step. 

## To Reproduce
Composer Home +New 
Select Enterprise assistant template
After project is loaded, add LUIS key and QnAMaker key
Start all 3 bots locally successfully
Test Enterprise Assistant in Emulator

## Expected behavior
User sees a more actionable message and knows what to do for this unauthorized error 

## Screenshots
![image](https://user-images.githubusercontent.com/25210203/115825876-e40b6d80-a3be-11eb-8d97-50873128784d.png)

lauren-mills:
Related to https://github.com/microsoft/BotFramework-Composer/issues/7255

lauren-mills:
I'm adding an error message if the allowedCallers isn't set. This should help with this issue

PR #991, opened 2021-04-23
[EA] Added allowed callers error message

lauren-mills:
If the root bot does not have allowedCallers configured and throws an error, the bot will send a message to explain.

This will only be shown in the emulator channel.

Fixes: #973

### DIFF
--- a/generators/generator-bot-enterprise-assistant/generators/app/templates/botName.dialog
+++ b/generators/generator-bot-enterprise-assistant/generators/app/templates/botName.dialog
@@ -335,7 +335,7 @@
                   "$designer": {
                     "id": "QVRnBT"
                   },
-                  "condition": "=exists(settings.skill) && not(exists(settings.runtimeSettings.skills.allowedCallers))",
+                  "condition": "=exists(settings.skill) && (not(exists(settings.runtimeSettings.skills.allowedCallers)) || count(where(settings.runtimeSettings.skills.allowedCallers, x, x != "")) == 0)",
                   "actions": [
                     {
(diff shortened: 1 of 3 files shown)


generators/generator-bot-enterprise-assistant/generators/app/templates/botName.dialog
```diff
@@ -330,6 +330,22 @@
             {
               "value": "emulator",
               "actions": [
+                {
+                  "$kind": "Microsoft.IfCondition",
+                  "$designer": {
+                    "id": "QVRnBT"
+                  },
+                  "condition": "=exists(settings.skill) && not(exists(settings.runtimeSettings.skills.allowedCallers))",
```
thomas-chung, line 338: you may also want to check for "" as well. I think if you have something like 

"skills" : { "allowedCallers" : [ "" ] } 

like they might have in a json template it would bypass this still.

thomas-chung: Otherwise this looks good.

thomas-chung (approved):

ryanlengel:
Should we pull this into Core Assistant as well? While it doesn't have preconfigured skills the intention is that you'd add your own skill and the same issue would arise.

lauren-mills:
> Should we pull this into Core Assistant as well? While it doesn't have preconfigured skills the intention is that you'd add your own skill and the same issue would arise.

@ryanlengel - yes, we should. I can add it in to this PR

thomas-chung (approved):

PR #992, opened 2021-04-23
chore: Component packages: Add SDK analyzers and common rules to maintain cross-component consistency

### LINKED ISSUE #982: Add SDK StyleCop rules to Component Repo

carlosscastro:
Fixes #982


tracyboehrer (approved):

packages/AdaptiveCards/AdaptiveCardsBotComponent.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.Bot.Components.AdaptiveCards
     /// <summary>
     /// <see cref="BotComponent"/> implementation for the adaptve card types.
     /// </summary>
-    public class AdaptiveBotComponent : BotComponent
+    public class AdaptiveCardsBotComponent : BotComponent
```
carlosscastro, line 16: I might need to revert this change depending on timing for this component. If it shipped or ships before this PR gets in, then I'll just revert the rename. Too bad that the file name is correct but the class name does not match it :(

packages/AdaptiveCards/AdaptiveCardComponentRegistration.cs
```diff
@@ -1,15 +1,17 @@
...
 
+using System;
 using Microsoft.Bot.Builder.Dialogs.Declarative.Obsolete;
 
 namespace Microsoft.Bot.Components.AdaptiveCards
 {
     /// <summary>
     /// <see cref="AdaptiveCardComponentRegistration"/> implementation for legacy runtimes.
     /// </summary>
+    [Obsolete("AdaptiveCardBotComponent is the new component definition.")]
```
peterinnesmsft, line 12: Missing 's' at the end of Card.

peterinnesmsft (approved):
:shipit:

lauren-mills (approved):

packages/Graph/Actions/Photo/GetPhoto.cs
```diff
@@ -61,7 +61,7 @@ protected override void PopulateParameters(DialogStateManager state, Dictionary<
         {
             if (this.UserId == null)
             {
-                throw new ArgumentNullException(nameof(this.UserId));
+                throw new InvalidOperationException($"GetPhoto requires UserId property populated.");
```
thomas-chung, line 64: Why can't we keep the ArgumentNullException but use the two string signature to point out the param name and keep the prettier message? I just think ArgumentNullException seems like a more descriptive exception overall.

thomas-chung (approved):
Besides the argumentnullexception -> invalidoperationexception change (which itself itsn't the end of the world). I think it looks good. Thanks Carlos!

packages/Graph/Actions/Photo/GetPhoto.cs
carlosscastro: @thomas-chung because ArgumentNullException is for arguments. There is no argument here being passed and the compiler knows it. The idea is to be consistent with the SDK and the Microsoft-wide SDK design guidelines which recommends most of these rules. Thinking it from a user's perspective, i.e. someone who is using this action, they are not invoking a method with this as argument -- they are filling a form through composer in most cases (i.e. in all composer scenarios, there is also authoring json and coded adaptive which is not the recommended code-first approach), and this exception means that they did not configure their action correctly, which feels more like a ConfigurationException, InvalidOperationException or ApplicationException.

We can choose to ignore these rules where it applies as long as it is reasonable, so it doesn't mean that they are carved in stone. In this case, I don't feel that this is one of those situations where we need to turn rules off. Instead of turning the general recommendation off, could we consider how do we change these exceptions' text so that it's nice and actionable and whether there is another type of exception that we'd prefer?

packages/AdaptiveCards/AdaptiveCardComponentRegistration.cs
carlosscastro: Good catch!

packages/Graph/Actions/Photo/GetPhoto.cs
thomas-chung: Got it, that makes sense, thanks! In this case I think having an InvalidOperationException sounds like a good candidate.

PR #1020, opened 2021-04-27
fix: Removed deploy.ps1 from adaptive generator

### LINKED ISSUE #952: Deploy.ps1 is out of date for the R13 runtime folder structure
The [scripts/deploy.ps1](https://github.com/microsoft/botframework-components/blob/main/generators/generator-bot-adaptive/generators/app/templates/dotnet/webapp/Scripts/deploy.ps1) script in the composer project is using the old runtime structure and needs to be updated to work with the new folder structure. 

Note, if this doesn't get addressed for R13 we will need to remove the scripts folder from the template (it doesn't currently work with the current folder structure), I think this script is referenced in docs so we will need to address that too if we delete it

tracyboehrer:
Fixes #952


gabog (changes requested):
I would remove the entire scripts folder, I don't anything there is up to date

tracyboehrer:
> I would remove the entire scripts folder, I don't anything there is up to date

I actually think it will.  When I look at the tree in the commit that folder is gone.  If not, I'll do a new PR to remove it.

PR #1023, opened 2021-04-27
fix: Add actionable responses to suggestions in Basic Assistant's bot tour

### LINKED ISSUE #1000: Basic assistant - Fails to understand query that the bot recommends to try
## Describe the bug

Setup and run the basic assistant. (LUIS Keys, QnA Key, etc). 
Click the card. 
image

Then click the skill:
image

None of the recommended text works:
image

Given that I've setup LUIS and QnA as needed, I expected these to work:
image

## Version

Version:  1.4.0-nightly.238698.3e4162
Electron: 8.2.4
Chrome: 80.0.3987.165
NodeJS: 12.13.0
V8: 8.0.426.27-electron.0

ryanlengel:
### Purpose
Close #1000
This updates the Bot Tour sample dialog in Basic Assistant to use dummy suggested actions that link to content the user can opt in to update. The previous activity is replaced with a multi choice prompt displaying the same Adaptive Card.

### Changes
#### Dialog
image

#### Expected flow
image
image

#### Interruptible
image


generators/generator-bot-core-assistant/generators/app/templates/dialogs/BotTourDialog/BotTourDialog.dialog
```diff
@@ -34,17 +34,79 @@
...
+          "unrecognizedPrompt": "",
+          "invalidPrompt": "",
+          "prompt": "${ChoiceInput_Prompt_BotTourDetails()}",
+          "choiceOptions": {
+            "includeNumbers": true,
+            "inlineOrMore": ", or ",
+            "inlineOr": " or ",
+            "inlineSeparator": ", "
+          },
+          "choices": [
```
peterinnesmsft, line 55: I trust this is valid, but it seems a little duplicative since the actions in the returned card already contain the results per se. Am I thinking about this wrong?

peterinnesmsft (approved):

generators/generator-bot-core-assistant/generators/app/templates/dialogs/BotTourDialog/BotTourDialog.dialog
ryanlengel: The original "prompt" was just a SendActivity with the card, where the card actions where separate intents (this logic remains in how we go from the Overview -> Detail card). This version replaces that logic with a multi-choice prompt with the display of those choices hidden. Instead, this has the old adaptive card but the data returned is a string that matches the expected choices listed here.

ryanlengel: I prefer this approach to handling different intents because in this case, the intents are just dummy data and there is no point in populating a new bot with a more complicated setup that they will just want to remove anyways.

generators/generator-bot-core-assistant/generators/app/templates/dialogs/BotTourDialog/language-generation/en-us/BotTourDialog.en-us.lg
```diff
@@ -188,34 +203,27 @@
...
+            "data": "B"
         },
         {
             "type": "Action.Submit",
-            "title": "\"Set a timer\"",
-            "data": {
-                "intent": "SampleAction"
-            }
+            "title": "\"Suggested action C\"",
+            "data": "C"
```
thomas-chung, line 225: _Ideally_ I'd prefer the Action.Submit would trigger an intent that would go to another dialog which is more realistic instead of triggering a switch statement. 

I think this is okay as a first pass to go. I think not having it going through to the unknown intent is already a lot better. Thanks Ryan!

thomas-chung (approved):

lauren-mills (approved):

generators/generator-bot-core-assistant/generators/app/templates/dialogs/BotTourDialog/language-generation/en-us/BotTourDialog.en-us.lg
ryanlengel: I agree, we can revisit the strategy for Bot Tour in R14. Thanks!

PR #1033, opened 2021-04-28
fix: replace Basic Assistant bot tour multichoice logic with intents

### LINKED ISSUE #1032: Basic Assistant: Interruption logic is not working as expected with latest BotTour changes
After additional testing I've discovered some issues with the multi-choice prompt introduced in #1023, namely:
- When used in an interruption of Bot Tour, chit-chat doesn't trigger as expected.
- After interrupting, the activity reprompts because the Bot Tour dialog is still on the stack. 

I'm going to use the previous SendActivity->EndDialog logic and have the sample utterances be handled by mock intents instead (the same pattern used before).

ryanlengel:
### Purpose
Close #1032 

### Changes
#1023 introduced additional edge cases where interruptions didn't work as expected, this addresses that by reverting to the better understood intent routing logic used before.

DetailsCard sends an activity with a card attachment.
image

Each suggested action maps to its own intent and response that can be customized by the user.
image

#### Expected flow
image

#### Interruptions
image
image

### DIFF
--- a/generators/generator-bot-core-assistant/generators/app/templates/dialogs/BotTourDialog/BotTourDialog.dialog
+++ b/generators/generator-bot-core-assistant/generators/app/templates/dialogs/BotTourDialog/BotTourDialog.dialog
@@ -34,79 +34,62 @@
       "intent": "DetailsCard",
       "actions": [
         {
-          "$kind": "Microsoft.ChoiceInput",
+          "$kind": "Microsoft.SendActivity",
           "$designer": {
-            "id": "8kK1VI"
+            "id": "ZMzFeP"
           },
-          "defaultLocale": "en-us",
-          "disabled": false,
-          "maxTurnCount": 3,
-          "alwaysPrompt": false,
-          "allowInterruptions": true,
-          "unrecognizedPrompt": "",
-          "invalidPrompt": "",
-          "prompt": "${ChoiceInput_Prompt_BotTourDetails()}",
-          "choiceOptions": {
-            "includeNumbers": true,
-            "inlineOrMore": ", or ",
-            "inlineOr": " or ",
-            "inlineSeparator": ", "
-          },
-          "choices": [
-            "A",
-            "B",
-            "C"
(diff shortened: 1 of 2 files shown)


generators/generator-bot-core-assistant/generators/app/templates/dialogs/BotTourDialog/BotTourDialog.dialog
```diff
@@ -34,79 +34,62 @@
...
+        }
+      ]
+    },
+    {
+      "$kind": "Microsoft.OnIntent",
+      "$designer": {
+        "id": "Fris5p",
+        "name": "ActionC"
+      },
+      "intent": "ActionC",
```
thomas-chung, line 85: Should there be dialog where the intent gets executed?

thomas-chung (changes requested):

generators/generator-bot-core-assistant/generators/app/templates/dialogs/BotTourDialog/BotTourDialog.dialog
ryanlengel: I'm not sure what you mean... this has a $kind of Microsoft.OnIntent, with an intent ("DetailsCard") and an array of actions on how to respond (in this case it just sends an activity). There doesn't need to be an additional reference to a dialog, it's in a dialog itself.

peterinnesmsft (approved):

generators/generator-bot-core-assistant/generators/app/templates/dialogs/BotTourDialog/BotTourDialog.dialog
thomas-chung: Ah got it, sorry I misread the json. :)

This is what happens when I am trying to visualize the json in the composer. :(

thomas-chung (approved):
Misread the json. Sorry!

PR #1066, opened 2021-05-03
chore: [Internal dev branch PR] Telephony: integration tests and CommandDialog interruption model first pass

carlosscastro:
#minor

### DIFF
--- a/packages/Telephony/Common/CommandDialog{T}.cs
+++ b/packages/Telephony/Common/CommandDialog{T}.cs
@@ -2,8 +2,6 @@
 // Licensed under the MIT License.
 
 using System;
-using System.Collections.Generic;
-using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using AdaptiveExpressions.Properties;
--- a/packages/Telephony/Common/CommandDialog{T}.cs
+++ b/packages/Telephony/Common/CommandDialog{T}.cs
@@ -9,7 +9,9 @@ using System.Threading.Tasks;
 using AdaptiveExpressions.Properties;
 using Microsoft.Bot.Builder;
 using Microsoft.Bot.Builder.Dialogs;
+using Microsoft.Bot.Builder.Dialogs.Adaptive;
 using Microsoft.Bot.Schema;
+using Newtonsoft.Json;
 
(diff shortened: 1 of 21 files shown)


packages/Telephony/Common/CommandDialog{T}.cs
```diff
@@ -54,30 +62,51 @@ public override async Task<DialogTurnResult> BeginDialogAsync(DialogContext dc,
...
-            // check activity
             var activity = dc.Context.Activity;
 
-            // If command result, handle it
+            // We are expecting a command result with the same name as our current CommandName.
             if (activity.Type == ActivityTypes.CommandResult
-                && activity.Name == this.Name.GetValue(dc.State))
+                && activity.Name == this.CommandName.GetValue(dc.State))
             {
                 // TODO: correlate command id before handling it.
```
p-nagpal, line 71: I think we can remove this TODO. Each command will decide if it needs correlation or is fire-and-forget.

packages/Telephony/Common/CommandDialog{T}.cs
```diff
@@ -54,30 +62,51 @@ public override async Task<DialogTurnResult> BeginDialogAsync(DialogContext dc,
...
+            // This activity was not the command result we were expecting. Mark as waiting and end the turn.
             return new DialogTurnResult(DialogTurnStatus.Waiting);
         }
 
+        /// <inheritdoc/>
+        protected override async Task<bool> OnPreBubbleEventAsync(DialogContext dc, DialogEvent e, CancellationToken cancellationToken)
+        {
+            if (e.Name == DialogEvents.ActivityReceived && dc.Context.Activity.Type == ActivityTypes.Message)
+            {
+                // Ask parent to perform recognition
```
p-nagpal, line 92: I noticed that the recognition block is before the interruption check. Will recognize utterance lead to activities being processed?

p-nagpal (approved):

packages/Telephony/Common/CommandDialog{T}.cs
carlosscastro: We are just emitting the event but if we return false in this method, then interruptions will not block this flow no matter the recognition.

PR #1140, opened 2021-06-04
fix: generator-bot-adaptive: Include *.model

### LINKED ISSUE #1129: Orchestrator: Need to add *.model files to Content Include List
[Composer#7892](https://github.com/microsoft/BotFramework-Composer/issues/7892): the multi-lingual model requires the *.model extensions to be allowed for publishing. 

## Version
1.0.0.GA

## Describe the bug
We need to allow more files through to allow the multi-lingual model to work correctly on Azure.  Add `**/*.model` to the **Content Include** list.

tracyboehrer:
Fixes #1129

### DIFF
--- a/generators/generator-bot-adaptive/generators/app/templates/dotnet/functions/botName.csproj
+++ b/generators/generator-bot-adaptive/generators/app/templates/dotnet/functions/botName.csproj
@@ -5,7 +5,7 @@
     <_FunctionsSkipCleanOutput>true</_FunctionsSkipCleanOutput>
   </PropertyGroup>
   <ItemGroup>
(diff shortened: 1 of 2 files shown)


peterinnesmsft:
I think we ought to evaluate utilization of build targets for handling this inclusion. Orchestrator should be fully empowered to tell us what files need to be included as content without us having to update the base generator each time.

tracyboehrer:
I suggest we merge this if this is the appropriate solution, and open an issue for the broader issue.  Orchestrator won't be the last component that requires additional files included, so I would be better to solve this sooner than later.

taicchoumsft (approved):

PR #1165, opened 2021-06-25
chore: Add tests for pause and resume recording

p-nagpal:
### Purpose
Complete test coverage

### Changes
1. Add tests for Pause and resume recording
2. Remove RecordingSettings as this is not supported but the channel anymore

### Tests
New tests have been added 

### Feature Plan
n/a

#minor


tests/packages/Microsoft.Bot.Components.Telephony.Tests/Integration/RecordingTests/StartRecording_HappyPath.test.dialog
```diff
@@ -7,22 +7,71 @@
             "$kind": "Microsoft.Test.UserSays",
             "text": "Hello I'm Calculon"
         },
+        {
+            "$kind": "Microsoft.Test.AssertReplyActivity",
```
p-nagpal, line 11: I have left the file names unchanged to allow for easy diff. Will rename to Recording_HappyPath.. in a next iteration.

packages/Telephony/Actions/StartRecording.cs
```diff
@@ -38,12 +37,6 @@ public StartRecording([CallerFilePath] string sourceFilePath = "", [CallerLineNu
             this.RegisterSourceLocation(sourceFilePath, sourceLineNumber);
 
             this.CommandName = RecordingStart;
-
-            this.Data = new RecordingStartSettings()
```
carlosscastro, line 42: Is this not relevant anymore?

carlosscastro (approved):

packages/Telephony/Actions/StartRecording.cs
p-nagpal: yeah this was removed because we only support a single value for the settings right now. This will come back as we support more settings.

Issue #1139, opened 2021-06-03
Teams - meeting start/end events (package)


clearab:
Add triggers for responding to meeting start/end events.

See [parent](https://github.com/microsoft/botframework-sdk/issues/6305)

Issue #1138, opened 2021-06-03
Teams - get meeting details (package)


clearab:
Add action to support calling new get meeting details API

See [parent](https://github.com/microsoft/botframework-sdk/issues/6304)

PR #1175, opened 2021-06-29
port: GetMeetingInfo, meeting start/end events [JS]

mdrichardson:
Addresses JS: #1138 
Addresses JS: #1139

#minor (have to include since this doesn't fully fix an issue)

### Purpose

Add the action, `GetMeetingInfo()`
Adds the Triggers, `OnTeamsMeetingStart`() and `OnTeamsMeetingEnd()`

### Tests

Note: The JS tests are a little tougher to verify by image. I stored the results of these methods in properties like `dialog.result` and then would "Send Response" `${dialog.result}`, but I don't think JS handles that well without errors. So, instead, would "Send Response" `${dialog.result.<some nested property that only is in the proper return object>`

I triggered `GetMeetingInfo()` to send on Unknown Intent (this is `result.details.id`, the Meeting ID):

image

Meeting start/end events (data is on `activity.value`, but this is `activity.value.startTime` or `...endTime`, as appropriate):

image


joshgummersall:
We need to update `peerDependencies` and `devDependencies` as well to account for needing newer SDK packages.

mdrichardson:
> We need to update `peerDependencies` and `devDependencies` as well to account for needing newer SDK packages.

@joshgummersall Good callout. My assumption was that that might happen in a different PR. However, this PR definitely depends on those changes. Any thoughts on to what version to bump to? Are we waiting until R14 GA?

joshgummersall:
For now you can use `~4.14.0-rc0` (including `preview` where applicable) which should match the remaining RCs as well. When we GA, we'll have to do one more bump to remove the `rcN` bit.

mdrichardson:
> For now you can use `~4.14.0-rc0` (including `preview` where applicable) which should match the remaining RCs as well. When we GA, we'll have to do one more bump to remove the `rcN` bit.

I forgot. Actually, since this is meant for composer, it can't actually be merged until at least RC1

joshgummersall:
Ah, fair enough. Let's get the other PR merged and some RCs built then!

mdrichardson:
Tested and working with RC1

image

packages/Teams/js/package.json
```diff
@@ -29,27 +29,27 @@
...
     "@typescript-eslint/eslint-plugin": "latest",
     "@typescript-eslint/parser": "latest",
-    "adaptive-expressions": "4.14.0-rc0",
-    "botbuilder": "4.14.0-rc0",
-    "botbuilder-dialogs": "4.14.0-rc0",
-    "botbuilder-dialogs-adaptive": "~4.14.0-preview.rc0",
-    "botbuilder-dialogs-adaptive-runtime-core": "~4.14.0-preview.rc0",
-    "botbuilder-dialogs-declarative": "~4.14.0-preview.rc0",
-    "botbuilder-stdlib": "~4.14.0-internal.rc0",
+    "adaptive-expressions": "~4.14.0-rc1",
```
joshgummersall, line 46: Technically (not a major issue since these are dev deps only) these should all lock to a specific version without a semver operator. We rely on `instanceof` in a few places which can break when transitive versions differ slightly.

joshgummersall (approved):
Small note about `devDependencies`, otherwise LGTM!

packages/Teams/js/package.json
mdrichardson: To clarify, lock all of the bot-related `devDependencies`? How about `peerDependencies`?

joshgummersall: Just dev deps. We want to allow users to use any compatible versions, and we assume they can ensure the compatible version they select is consistent across all SDK packages.

PR #1181, opened 2021-06-30
IGNORE THIS : chore: call transfer tests and uischema update

vanshu86:
### Purpose

call transfer tests and uischema update

### Changes

no

### Tests

yes


packages/Telephony/Schemas/Microsoft.Telephony.CallTransfer.uischema
```diff
@@ -1,7 +1,7 @@
 ﻿{
   "$schema": "https://schemas.botframework.com/schemas/ui/v1.0/ui.schema",
   "form": {
-    "label": "Transfer a call",
+    "label": "Call Transfer",
     "subtitle": "Call Transfer",
```
carlosscastro, line 5: Please add a more informative subtitle. Composer shows the label in large, bold letters and the subtitle right below with lighter letter. Adding a longer sentence will much improve the user experience. Example that is not necessarily the content we want but an instance of a longer informative sentence: `Transfers a call to the provided phone number when using your bot in Telephony.`

tests/packages/Microsoft.Bot.Components.Telephony.Tests/Integration/CallTransferTests/CallTransfer_BaseScenario.test.dialog
```diff
@@ -5,12 +5,16 @@
...
-                    "phoneNumber": "+1555443443"
-                }
-            ]
+          "actions": [
+            {
+              "$kind": "Microsoft.Telephony.CallTransfer",
+              "phoneNumber": "+1555443443"
+            },
+            {
+              "$kind": "Microsoft.SendActivity",
```
p-nagpal, line 14: In a real TC call, will we receive this activity? Or will this activity be dropped and the call terminate with an EndOfConversation activity?

tests/packages/Microsoft.Bot.Components.Telephony.Tests/Integration/CallTransferTests/CallTransfer_HappyPath.test.dialog
```diff
@@ -2,17 +2,21 @@
...
-        },
-        {
-            "$kind": "Microsoft.Test.AssertReplyActivity",
-            "assertions": [
-                "type == 'event'",
-                "value.TargetPhoneNumber == '+1555443443'"
-            ]
-        }
-    ]
+  "script": [
```
p-nagpal, line 5: Minor - will be nice to have tabs / space aligned.

tests/packages/Microsoft.Bot.Components.Telephony.Tests/Integration/CallTransferTests/CallTransfer_BaseScenario.test.dialog
```diff
@@ -5,12 +5,16 @@
...
-            "actions": [
-                {
-                    "$kind": "Microsoft.Telephony.CallTransfer",
-                    "phoneNumber": "+1555443443"
-                }
-            ]
+          "actions": [
+            {
+              "$kind": "Microsoft.Telephony.CallTransfer",
+              "phoneNumber": "+1555443443"
```
p-nagpal, line 11: Regding negative test for invalid or missing number: phoneNumber needs to be marked as mandatory or validated in the action. 

    var phoneNumber = this.PhoneNumber?.GetValue(dc.State); // accepts empty and invalid numbers

p-nagpal (changes requested):

tests/packages/Microsoft.Bot.Components.Telephony.Tests/Integration/CallTransferTests/CallTransfer_IgnoredInNonTelephonyChannel.test.dialog
```diff
@@ -0,0 +1,11 @@
+{
+  "$schema": "../../../tests.schema",
```
carlosscastro, line 2: Shouldn't you verify that the action after the call transfer gets executed here? The action before the call transfer will always get executed in all channels, so this test will be tautological.

thomas-chung:
@vanshu86 can you please clarify if this and the associated feature branch need to go in for R14?

carlosscastro:
@thomas-chung  this branch is to be merged to main post r14

thomas-chung:
> @thomas-chung this branch is to be merged to main post r14

Thanks for clarifying! 👯‍♂️

tests/packages/Microsoft.Bot.Components.Telephony.Tests/Integration/CallTransferTests/CallTransfer_IgnoredInNonTelephonyChannel.test.dialog
vanshu86: From what I understood, that call transfer should not work on non telephony channel, so should I be expecting something here? perhaps I should add a AssertNoActivity?

tests/packages/Microsoft.Bot.Components.Telephony.Tests/Integration/CallTransferTests/CallTransfer_BaseScenario.test.dialog
vanshu86: I have added validation to the phone number

tests/packages/Microsoft.Bot.Components.Telephony.Tests/Integration/CallTransferTests/CallTransfer_BaseScenario.test.dialog
vanshu86: Good question, I think in real call once we finish call transfer we cant get back to bot control (you might know better :)). I can remove this sendactivity

tests/packages/Microsoft.Bot.Components.Telephony.Tests/Integration/CallTransferTests/CallTransfer_IgnoredInNonTelephonyChannel.test.dialog
carlosscastro: What is your design for the behavior in another channels? I'd assume that it is not that 'it should not work' and rather what we did with start recording, which is that in other channels, the call transfer is a no-op and we continue with the next action. If you don't have a design for this we can discuss in the telephony crew chat.

tests/packages/Microsoft.Bot.Components.Telephony.Tests/Integration/CallTransferTests/CallTransfer_BaseScenario.test.dialog
carlosscastro: I think the posterior send activity is good, because in telephony channel we should never see this one (since it ended up getting transferred) and in non-telephony channels we should see this (because it should be a no-op)

tests/packages/Microsoft.Bot.Components.Telephony.Tests/Integration/CallTransferTests/CallTransfer_BaseScenario.test.dialog
carlosscastro: I understood from the chat that we'd validate non-null but we'd leave specific validation to the channels so we don't hardcode provider specific validation. I also understood that if we pass a non-null value along with the event, we'd get an error from the channel if the number is not valid. Are we saying that we'll end up validating 1) non-null and 2) phone regex  in the action then?

Regarding getting the error back on bad phone numbers, which was discussed yesterday, have you tested it with the actual telephony channel? How do we get the error back? do we get it as response to the event activity or how? We should make sure that we have a solid story here in capturing such errors.

vanshu86: Please look at my latest PR for the error 

var error = "Phone number is empty, please provide number in E.164 format";
 return await dc.EndDialogAsync(result: error, cancellationToken: cancellationToken).ConfigureAwait(false);

this error will be caught in runtime, when actual phone call will be made via telephony and call transfer is initiated, so not sure where we should bubble this up per say.

tests/packages/Microsoft.Bot.Components.Telephony.Tests/Integration/CallTransferTests/CallTransfer_BaseScenario.test.dialog
vanshu86: we do get control of the bot in this test framework, even in telephony channel, since we dont do the actual transfer from what I can see

tests/packages/Microsoft.Bot.Components.Telephony.Tests/Integration/CallTransferTests/CallTransfer_BaseScenario.test.dialog
carlosscastro: I see that you added the error, my comment was about whether to perform validation in the bot side or in the channel / provider side as we discussed in the crew. Let's talk this in a call.

packages/Telephony/Actions/CallTransfer.cs
```diff
@@ -55,12 +56,24 @@ public CallTransfer([CallerFilePath] string sourceFilePath = "", [CallerLineNumb
         /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
         public async override Task<DialogTurnResult> BeginDialogAsync(DialogContext dc, object options = null, CancellationToken cancellationToken = default(CancellationToken))
         {
-            var phoneNumber = this.PhoneNumber?.GetValue(dc.State);
+            if (dc.Context.Activity.ChannelId == Channels.Telephony)
+            {
+                var error = "Phone number is empty, please provide number in E.164 format";
+                var regexFormat = "^\\+\\d{1,3}\\s\\d{1,14}(\\s\\d{1,13})?";
```
p-nagpal, line 62: I cannot tell if this regex meets E.164  standard :) Likely we might miss something: https://www.itu.int/rec/T-REC-E.164-201011-I/en

packages/Telephony/Actions/CallTransfer.cs
```diff
@@ -55,12 +56,24 @@ public CallTransfer([CallerFilePath] string sourceFilePath = "", [CallerLineNumb
...
 
-            // Create handoff event, passing the phone number to transfer to as context.
-            HandoffContext = new { TargetPhoneNumber = phoneNumber };
-            
-            return await base.BeginDialogAsync(dc, options, cancellationToken).ConfigureAwait(false);
+                var phoneNumber = this.PhoneNumber?.GetValue(dc.State);
+
+                if (string.IsNullOrEmpty(phoneNumber) || Regex.IsMatch(phoneNumber, regexFormat))
```
p-nagpal, line 66: Do you mean &&

tests/packages/Microsoft.Bot.Components.Telephony.Tests/Integration/CallTransferTests/CallTransfer_BaseScenario.test.dialog
vanshu86: yes, I spoke to TC team and they dont do validation of phonenumber on their side, they rely on validation on our side else it will fail in the lowest level, which could be detected much earlier by us, hence I am adding validation here.

tests/packages/Microsoft.Bot.Components.Telephony.Tests/Integration/CallTransferTests/CallTransfer_IgnoredInNonTelephonyChannel.test.dialog
vanshu86: As discussed offline, other channels should not support call transfer - Pr has also been updated accordingly

packages/Telephony/Actions/CallTransfer.cs
carlosscastro: This regex is not right, the code below returns `false` when I just tested it in vacuum using your test phone number

```csharp
	public static void Main()
	{
		var regexFormat = "^\\+\\d{1,3}\\s\\d{1,14}(\\s\\d{1,13})?";
		Console.WriteLine(Regex.IsMatch("+1555443443", regexFormat));
		
	}
```

Regardless, let's discuss on Tuesday when we go through the spec since I'm not sure that regex validation is a good thing to do here. And if we do want it, we can discuss the trade offs of different approaches together in a call which is more efficient.

packages/Telephony/Actions/CallTransfer.cs
carlosscastro: Bit more detail: This condition does not work as-is, since if the regex is a match (i.e. good phone number) we'll send an error and never transfer the call. Something like `if (string.IsNullOrEmpty(phoneNumber) && !Regex.IsMatch(phoneNumber, regexFormat))` could yield the desired results. However again, let's discuss whether we want or not to include the validation on Tuesday.

You might wonder, if you did run the tests, why did they pass? The reason your test could have passed if you ran them is that 1) the regex is wrong , so it will never return true, and 2) the tests you added are all for non-telephony channel (which you specified in the cs explicitly ) so the behavior will not exercise this code

tests/packages/Microsoft.Bot.Components.Telephony.Tests/CallTransferTests.cs
```diff
@@ -21,5 +21,23 @@ public async Task CallTransfer_HappyPath()
...
+        [Fact]
+        public async Task CallTransfer_EmptyPhoneNumber()
+        {
+            await TestUtils.RunTestScript(_resourceExplorerFixture.ResourceExplorer, adapterChannel: Channels.Msteams);
+        }
+
+        [Fact]
+        public async Task CallTransfer_IncorrectPhoneNumber()
+        {
+            await TestUtils.RunTestScript(_resourceExplorerFixture.ResourceExplorer, adapterChannel: Channels.Msteams);
```
carlosscastro, line 34: You probably don't want teams channel here

tests/packages/Microsoft.Bot.Components.Telephony.Tests/CallTransferTests.cs
```diff
@@ -21,5 +21,23 @@ public async Task CallTransfer_HappyPath()
         {
             await TestUtils.RunTestScript(_resourceExplorerFixture.ResourceExplorer, adapterChannel: Channels.Telephony);
         }
+
+        [Fact]
+        public async Task CallTransfer_EmptyPhoneNumber()
+        {
+            await TestUtils.RunTestScript(_resourceExplorerFixture.ResourceExplorer, adapterChannel: Channels.Msteams);
```
carlosscastro, line 28: You probably don't want teams channel here

tests/packages/Microsoft.Bot.Components.Telephony.Tests/Integration/CallTransferTests/CallTransfer_BaseScenario.test.dialog
carlosscastro: Thanks! Let's discuss this with Pooja on Tuesday.

vanshu86: discussed with Pooja offline, we will not do validation for format. But we need your input on how to localize error for null case

packages/Telephony/Actions/CallTransfer.cs
vanshu86: thanks, this is the issue with not using standard validation here, removing this validation as we agreed upon

packages/Telephony/Actions/CallTransfer.cs
vanshu86: removing this validation

tests/packages/Microsoft.Bot.Components.Telephony.Tests/CallTransferTests.cs
vanshu86: yes my bad, copy paste error

packages/Telephony/Actions/CallTransfer.cs
```diff
@@ -2,12 +2,13 @@
 // Licensed under the MIT License.
 
 using System.Runtime.CompilerServices;
+using System.Text.RegularExpressions;
```
vanshu86, line 5: need to remove

PR #1193, opened 2021-07-07
feat: DTMF Batching for Telephony package

goldenwitch:
### Purpose
To resolve issue https://github.com/microsoft/botframework-telephony/issues/46

### Changes
Added action for DTMF batching. Supports both termination characters and fixed length batches.

### Tests

Added happy path tests for new action.

### DIFF
--- a/packages/Telephony/Actions/BatchDTMF.cs
+++ b/packages/Telephony/Actions/BatchDTMF.cs
@@ -1,0 +1,109 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Runtime.CompilerServices;
+using System.Text.RegularExpressions;
+using System.Threading;
+using System.Threading.Tasks;
+
+using AdaptiveExpressions.Properties;
+using Microsoft.Bot.Builder.Dialogs;
+using Newtonsoft.Json;
+
+namespace Microsoft.Bot.Components.Telephony.Actions
+{
+    /// <summary>
+    /// Aggregates input until it matches a regex pattern and then stores the result in an output property
+    /// </summary>
+    public class BatchDTMF : Dialog
+    {
+        /// <summary>
+        /// Class identifier.
+        /// </summary>
+        [JsonProperty("$kind")]
+        public const string Kind = "Microsoft.Telephony.BatchDTMF";
+        private const string AggregationDialogMemory = "dialog.aggregation";
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="BatchDTMF"/> class.
+        /// </summary>
+        /// <param name="sourceFilePath">Optional, source file full path.</param>
+        /// <param name="sourceLineNumber">Optional, line number in source file.</param>
+        [JsonConstructor]
+        public BatchDTMF([CallerFilePath] string sourceFilePath = "", [CallerLineNumber] int sourceLineNumber = 0)
+            : base()
+        {
+            // enable instances of this command as debug break point
+            this.RegisterSourceLocation(sourceFilePath, sourceLineNumber);
+        }
+
+        /// <summary>
+        /// Gets or sets the regex pattern to use to decide when the dialog has aggregated the whole message.
+        /// </summary>
+        [JsonProperty("terminationConditionRegexPattern")]
+        public StringExpression TerminationConditionRegexPattern { get; set; }
+
+        /// <summary>
+        /// Gets or sets the property to assign the result to.
+        /// </summary>
+        [JsonProperty("property")]
+        public StringExpression Property { get; set; }
+
+        /// <inheritdoc/>
+        public override async Task<DialogTurnResult> BeginDialogAsync(DialogContext dc, object options = null, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            //Get value of temrination string from expression
+            string regexPattern = this.TerminationConditionRegexPattern?.GetValue(dc.State);
+
+            if (dc.Context.Activity.Type != "message")
+            {
+                return new DialogTurnResult(DialogTurnStatus.Waiting);
+            }
+
+            //Is the current message the termination string?
+            if (new Regex(regexPattern).IsMatch(dc.Context.Activity.Text))
+            {
+                //If so, store the message in the property specified and end this dialog
+                //Get property from expression
+                string property = this.Property?.GetValue(dc.State);
+
+                dc.State.SetValue(property, dc.Context.Activity.Text);
+                return await dc.EndDialogAsync(cancellationToken: cancellationToken).ConfigureAwait(false);
+            }
+            else
+            {
+                //else, append the message to an aggregation memory state and end the turn
+                dc.State.SetValue(AggregationDialogMemory, dc.Context.Activity.Text);
+                return new DialogTurnResult(DialogTurnStatus.Waiting);
+            }
+        }
+
+        /// <inheritdoc/>
+        public override async Task<DialogTurnResult> ContinueDialogAsync(DialogContext dc, CancellationToken cancellationToken = default)
+        {
+            //Get value of temrination string from expression
+            string regexPattern = this.TerminationConditionRegexPattern?.GetValue(dc.State);
+
+            //append the message to the aggregation memory state
+            var existingAggregation = dc.State.GetValue(AggregationDialogMemory, () => string.Empty);
+            existingAggregation += dc.Context.Activity.Text;
+
+            //Is the current aggregated message the termination string?
+            if (new Regex(regexPattern).IsMatch(existingAggregation))
+            {
+                //If so, save it to the output property and end the dialog
+                //Get property from expression
+                string property = this.Property?.GetValue(dc.State);
+
+                dc.State.SetValue(property, existingAggregation);
+                return await dc.EndDialogAsync(cancellationToken: cancellationToken).ConfigureAwait(false);
+            }
+            else
+            {
+                //else, save the updated aggregation and end the turn
+                dc.State.SetValue(AggregationDialogMemory, existingAggregation);
+                return new DialogTurnResult(DialogTurnStatus.Waiting);
+            }
+        }
+    }
+}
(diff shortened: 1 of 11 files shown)


packages/Telephony/Actions/BatchDTMF.cs
```diff
@@ -0,0 +1,109 @@
...
+        [JsonConstructor]
+        public BatchDTMF([CallerFilePath] string sourceFilePath = "", [CallerLineNumber] int sourceLineNumber = 0)
+            : base()
+        {
+            // enable instances of this command as debug break point
+            this.RegisterSourceLocation(sourceFilePath, sourceLineNumber);
+        }
+
+        /// <summary>
+        /// Gets or sets the regex pattern to use to decide when the dialog has aggregated the whole message.
```
arturl, line 41: The dialog is called DTMF but regex supports non-digits. Does it need to be that general?

packages/Telephony/Actions/BatchDTMF.cs
```diff
@@ -0,0 +1,109 @@
...
+using AdaptiveExpressions.Properties;
+using Microsoft.Bot.Builder.Dialogs;
+using Newtonsoft.Json;
+
+namespace Microsoft.Bot.Components.Telephony.Actions
+{
+    /// <summary>
+    /// Aggregates input until it matches a regex pattern and then stores the result in an output property
+    /// </summary>
+    public class BatchDTMF : Dialog
```
arturl, line 18: How do we combine voice and DTMF input in the prompt? ("Please say or type your credit card number"). Will this just work?

goldenwitch: Yes, as it is, it will just work.

packages/Telephony/Actions/BatchDTMF.cs
goldenwitch: It's probably a good idea to split this out into an underlying "aggregator dialog" and some stuff that builds on top of it with more specificity (as you suggested offline :D).

packages/Telephony/Actions/BatchDTMF.cs
```diff
@@ -0,0 +1,109 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Runtime.CompilerServices;
+using System.Text.RegularExpressions;
+using System.Threading;
+using System.Threading.Tasks;
+
```
carlosscastro, line 8: nit: remove space for consistency

carlosscastro: Nothing against the fact that you thought of this and separated System.* from the others, which I imagine you did to keep things neat and organized. However, better to do what all the other codebase does in this case.

packages/Telephony/Actions/BatchDTMF.cs
carlosscastro, line 18: Naming: I am not a fan of BatchDTMF, it's pretty hard to assess what this does except you are already deep in telephony.  If I'm a bot developer new to telephony world, I don't know that DTMF is and I don't know why it needs to be batched.

Could you brainstorm on more intuitive names? Also, when a dialog asks for a question and expects an answer with certain, we named the other similar dialogs `<WhatToGet>Input`.

Some brainstorming (please butcher, criticize and suggest others!)

- DialToneInput
- TelephonyInput
- DTMFInput
- BatchDTMFInput
- KeyPressInput
- MultiDialToneInput

I know some are very bad, but its a brainstorm!

packages/Telephony/Actions/BatchDTMF.cs
carlosscastro: Could you explain the aggregator dialog thought?

packages/Telephony/Actions/BatchDTMF.cs
```diff
@@ -0,0 +1,109 @@
...
+        {
+            // enable instances of this command as debug break point
+            this.RegisterSourceLocation(sourceFilePath, sourceLineNumber);
+        }
+
+        /// <summary>
+        /// Gets or sets the regex pattern to use to decide when the dialog has aggregated the whole message.
+        /// </summary>
+        [JsonProperty("terminationConditionRegexPattern")]
+        public StringExpression TerminationConditionRegexPattern { get; set; }
```
carlosscastro, line 44: Functionality / extensibility: We are locking in to regex here. We've talked about 3 kinds of termination:

1. Regex is fulfilled
2. Character count is met
3. Configurable terminator character

Would these 3 be desirable still? If yes, what's the future / extensibility story? would we ahve a single DTMF dialog for the 3 terminations or would we have an injectable terminator of some sort?

carlosscastro, line 44: Naming: would `TerminationConditionRegex` do the trick? Not sure if pattern adds a lot and makes the name pretty lengthy. Brainstorming here, would love to hear thoughts.

packages/Telephony/Actions/BatchDTMF.cs
carlosscastro, line 18: Would we get value deriving from InputDialog? (which has the retries and other stuff built in). Not sure if we would but worth checking.

packages/Telephony/Actions/BatchDTMF.cs
```diff
@@ -0,0 +1,109 @@
...
+        /// Gets or sets the property to assign the result to.
+        /// </summary>
+        [JsonProperty("property")]
+        public StringExpression Property { get; set; }
+
+        /// <inheritdoc/>
+        public override async Task<DialogTurnResult> BeginDialogAsync(DialogContext dc, object options = null, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            //Get value of temrination string from expression
+            string regexPattern = this.TerminationConditionRegexPattern?.GetValue(dc.State);
```
carlosscastro, line 56: Should this be required?

packages/Telephony/Actions/BatchDTMF.cs
```diff
@@ -0,0 +1,109 @@
...
+
+            if (dc.Context.Activity.Type != "message")
```
carlosscastro, line 58:
```suggestion
            if (dc.Context.Activity.Type != ActivityTypes.Message)
```

packages/Telephony/Actions/BatchDTMF.cs
```diff
@@ -0,0 +1,109 @@
...
+            {
+                return new DialogTurnResult(DialogTurnStatus.Waiting);
+            }
+
+            //Is the current message the termination string?
+            if (new Regex(regexPattern).IsMatch(dc.Context.Activity.Text))
```
carlosscastro, line 64: This results in compiling regex every time which has a pretty intense perf implication depending on the pattern. The internal regex implementation has a cache. You cannot really keep the regex instance but what you can do is use the static overload, i.e. Regex.Match. From the[ regex docs](https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regex?view=net-5.0).

> The regular expression engine must compile a particular pattern before the pattern can be used. Because Regex objects are immutable, this is a one-time procedure that occurs when a Regex class constructor or a static method is called. To eliminate the need to repeatedly compile a single regular expression, the regular expression engine caches the compiled regular expressions used in static method calls. As a result, regular expression pattern-matching methods offer comparable performance for static and instance methods.

So, using the static overload you get the free caching to avoid recompilation.

Also, as a general rule you always want to pass a timeout to regex evaluation to avoid malicious inputs other RegexDenialOfService attacks. Unfortunately, I think the static overload does not have a timeout overload, if that is the case it's a 'choose your poison' scenario, but we can talk about it

packages/Telephony/Actions/BatchDTMF.cs
carlosscastro: Ha, I jsut realized that this is not a prompt, it's a dialog to read the info. Do we want to make it a prompt / input dialog? If not, then no need to derive from InputDialog. The thing is that if I ask for SSN and you miss the regex, InputDialog already has all the logic to retry, plus the fields to allow developers to pass templates on how to prompt, how to reprompt, how many retries and how to tell users you didnt understand. Worth considering, happy to help if you think that is a reasonable path.

packages/Telephony/Actions/BatchDTMF.cs
```diff
@@ -0,0 +1,109 @@
...
+                string property = this.Property?.GetValue(dc.State);
+
+                dc.State.SetValue(property, existingAggregation);
+                return await dc.EndDialogAsync(cancellationToken: cancellationToken).ConfigureAwait(false);
+            }
+            else
+            {
+                //else, save the updated aggregation and end the turn
+                dc.State.SetValue(AggregationDialogMemory, existingAggregation);
+                return new DialogTurnResult(DialogTurnStatus.Waiting);
```
carlosscastro, line 105: If my regex is '[1-3]*' and I press the 4 key endlessly, wouldn't this end in an infinite loop?

packages/Telephony/Actions/BatchDTMF.cs
carlosscastro: Of course they can all be done through regexes, however I'm thinking from a non-developer perspective, where 2. and 3. are muuuch more approachable.

carlosscastro: Brainstorm: Another approach is we could have the regex one be the coded action and then have 2 more actions in schemas that use the same class but when we load them from json, if they specify 'terminatorCharacter' property, in our converter we translate that to a well known regex that we set in the class.

carlosscastro (commented):
Thanks Autumn!!! Excited about this feature. Asking some questions and comments. The nit level comments can be deferred, mostly we should discuss 1) terminators and the fact that regex is too complicated potentially for non-developers and 2) whether this should be an input dialog instead of a regular dialog, so we benefit from retries, and a user experience around them (though fit would need to be assessed)

goldenwitch:
new PR at https://github.com/microsoft/botframework-components/pull/1219

PR #1198, opened 2021-07-07
chore: Fix deprecated Nuget pcakage elements. Fixes NU5048 and NU5125 warnings

p-nagpal:
### Purpose

Fixes the following Nuget warnings when generating the package:

NU5048 The 'PackageIconUrl'/'iconUrl' element is deprecated. Consider using the 'PackageIcon'/'icon' element instead. Learn more at https://aka.ms/deprecateIconUrl

NU5125 The 'licenseUrl' element will be deprecated. Consider using the 'license' element instead.

### Changes

Change project file according to docs below: 

https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5048?f1url=%3FappId%3DDev16IDEF1%26l%3DEN-US%26k%3Dk(NU5048)%26rd%3Dtrue

https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5125?f1url=%3FappId%3DDev16IDEF1%26l%3DEN-US%26k%3Dk(NU5125)%26rd%3Dtrue

#minor


carlosscastro (approved):

ryanlengel (changes requested):
For licenseUrl, I have a PR #1197 out that uses `<PackageLicenseExpression>MIT</PackageLicenseExpression>` rather than referencing a license file directly. This is the same that is used in the dotnet repo.

For PackageIconUrl/iconUrl, I believe that we need to continue using these as they reference a remote URL and this isn't supported with PackageIcon/icon. If we switch to use the recommended properties we should host the icon.png ourselves rather than pointing to an empty path (as the icon is used in Composer's Package Manager). Not sure if we want to risk having multiple copies of the Bot Builder icon.png rather than all point to the same location. The dotnet repo is still using the url, for reference.

p-nagpal:
> For licenseUrl, I have a PR #1197 out that uses `<PackageLicenseExpression>MIT</PackageLicenseExpression>` rather than referencing a license file directly. This is the same that is used in the dotnet repo.
> 
> For PackageIconUrl/iconUrl, I believe that we need to continue using these as they reference a remote URL and this isn't supported with PackageIcon/icon. If we switch to use the recommended properties we should host the icon.png ourselves rather than pointing to an empty path (as the icon is used in Composer's Package Manager). Not sure if we want to risk having multiple copies of the Bot Builder icon.png rather than all point to the same location. The dotnet repo is still using the url, for reference.

Thanks @ryanlengel . In that case, I can merge in your PR for License. For IconUrl, is it alright to ignore NU5048 since we cannot change the file location?

carlosscastro:
Thanks for the input @ryanlengel !!!!

ryanlengel:
> > For licenseUrl, I have a PR #1197 out that uses `<PackageLicenseExpression>MIT</PackageLicenseExpression>` rather than referencing a license file directly. This is the same that is used in the dotnet repo.
> > For PackageIconUrl/iconUrl, I believe that we need to continue using these as they reference a remote URL and this isn't supported with PackageIcon/icon. If we switch to use the recommended properties we should host the icon.png ourselves rather than pointing to an empty path (as the icon is used in Composer's Package Manager). Not sure if we want to risk having multiple copies of the Bot Builder icon.png rather than all point to the same location. The dotnet repo is still using the url, for reference.
> 
> Thanks @ryanlengel . In that case, I can merge in your PR for License. For IconUrl, is it alright to ignore NU5048 since we cannot change the file location?

Yes, I added NU5048 to NoWarn so it shouldn't warn us any more. PR merged, feel free to cherry-pick to your branch! :)

p-nagpal:
Abandoning this PR. Will merge 1197 from main instead.

PR #1236, opened 2021-07-23
feat: Interruption support for DTMF batching

goldenwitch:
### Purpose
resolves: microsoft/botframework-telephony#93 #minor
By adding support for interruption flows and interruption controls to the DTMF batching functionality

### Changes
Interruptions will now work. If the parent dialog recognizes an intent that the DTMF batching functionality doesn't handle, it will handle that functionality and return control to the DTMF batching dialog.

The DTMF batching dialog will then either reprompt or give up depending on user settings.

Also added a bunch of random UX fixes and renamed the kind for the dtmf batching input

### Tests
Covered by 4 new tests.


packages/Telephony/Actions/BatchTerminationCharacterInput.cs
```diff
@@ -68,5 +80,28 @@ public override Task<DialogTurnResult> ContinueDialogAsync(DialogContext dc, Can
                 return Task.FromResult(new DialogTurnResult(DialogTurnStatus.Waiting));
             }
         }
+
+        /// <inheritdoc/>
+        protected override async Task<bool> OnPreBubbleEventAsync(DialogContext dc, DialogEvent e, CancellationToken cancellationToken)
```
carlosscastro, line 85: Question: can we have OnPreBubbleEvent and AllowInterruption properties be in the base class so that we have that code only once?

packages/Telephony/Common/RegexAggregatorInput.cs
```diff
@@ -17,14 +18,14 @@ namespace Microsoft.Bot.Components.Telephony.Actions
     /// <summary>
     /// Aggregates input until it matches a regex pattern and then stores the result in an output property.
     /// </summary>
-    public class RegexAggregatorInput : Dialog
+    public abstract class RegexAggregatorInput : Dialog
```
carlosscastro, line 21: If this is abstract, that means that we don't expect it to be instantiable through declarative. In that case, can we make it internal, and then remove the Kind property?

packages/Telephony/Actions/BatchTerminationCharacterInput.cs
goldenwitch: Done

arturl (approved):

carlosscastro (approved):

PR #1242, opened 2021-08-09
chore: implemented interrupt mask for digits in DTMF batching actions

goldenwitch:
### Purpose

We think it's probably better if DTMF batching doesn't bubble digits by default, in case people are using top level intents for their menus. I implemented this.

### Changes

Added an "Interruption Mask" to the regex aggregator, and then utilized it in the DTMF actions

### Tests
Added two new tests to cover this case.

### DIFF
--- a/packages/Telephony/Actions/BatchFixedLengthInput.cs
+++ b/packages/Telephony/Actions/BatchFixedLengthInput.cs
@@ -22,6 +22,7 @@ namespace Microsoft.Bot.Components.Telephony.Actions
         public const string Kind = "Microsoft.Telephony.BatchFixedLengthInput";
 
         private const string _dtmfCharacterRegex = @"^[\d#\*]+$";
+        private const string _interruptionMaskRegex = @"^[\d]+$";
         private int _batchLength;
 
         /// <summary>
@@ -36,6 +37,7 @@ namespace Microsoft.Bot.Components.Telephony.Actions
         {
             // enable instances of this command as debug break point
             this.RegisterSourceLocation(sourceFilePath, sourceLineNumber);
+            InterruptionMask = _interruptionMaskRegex;
         }
 
         /// <summary>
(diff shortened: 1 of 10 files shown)


packages/Telephony/Actions/BatchTerminationCharacterInput.cs
```diff
@@ -22,6 +22,7 @@ public class BatchTerminationCharacterInput : RegexAggregatorInput
         public const string Kind = "Microsoft.Telephony.BatchTerminationCharacterInput";
 
         private const string _dtmfCharacterRegex = @"^[\d#\*]+$";
```
p-nagpal, line 24: Looks like we cannot change the termination character from #. TerminationConditionRegexPattern and _terminationCharacter are not used anywhere.

packages/Telephony/Actions/BatchFixedLengthInput.cs
```diff
@@ -22,6 +22,7 @@ public class BatchFixedLengthInput : RegexAggregatorInput
         public const string Kind = "Microsoft.Telephony.BatchFixedLengthInput";
 
         private const string _dtmfCharacterRegex = @"^[\d#\*]+$";
+        private const string _interruptionMaskRegex = @"^[\d]+$";
```
p-nagpal, line 25: Minor: add comment like "digits only regex" for easier reading of the code ;)

p-nagpal (approved):
Will be good to call out limitation on termination character (# only) and add it to the next version instead.

packages/Telephony/Actions/BatchTerminationCharacterInput.cs
goldenwitch: It actually works with any termination character (in the DTMF character set). I did some tricky bits with the property where it sets the parents property as part of the getter/setter (kind of a virtual property pattern).

packages/Telephony/Actions/BatchFixedLengthInput.cs
goldenwitch: smart.

packages/Telephony/Actions/BatchTerminationCharacterInput.cs
p-nagpal: Agreed. I misread the _dtmfCharacterRegex = @"^[\d#\*]+$" as something else. Def. worth adding more comments for the slow regex readers.

p-nagpal (approved):

PR #1293, opened 2022-01-27
fix: bug with globby not excluding files correctly

lauren-mills:
### Purpose
Fixes https://github.com/microsoft/BotFramework-Composer/issues/8927

### Changes
Globby dependency changed its pattern for excluding files which broken the generation of the .sln file for the Enterprise Assistant Template. This PR resolves that issue

### DIFF
--- a/generators/generator-bot-enterprise-assistant/generators/app/index.js
+++ b/generators/generator-bot-enterprise-assistant/generators/app/index.js
@@ -78,7 +78,7 @@ module.exports = class extends BaseGenerator {
 
   writing() {
     this._copyBotTemplateFiles({
-      path: ['**', '!(*.sln)'],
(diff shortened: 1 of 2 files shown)


ryanlengel (approved):

Markyboy944:
Hi there, any thoughts on timescales for this fix ? (I'm still getting the eror with latest composer v 2.1.1. , so I assume nothing released yet. many thanks

benbrown:
@Markyboy944 The process to test, validate and publish these changes is underway. You will not need to update Composer as the templates are independently versioned.  We'll post here when that happens.

benbrown:
@Markyboy944 This is all resolved with the new 1.2.2 templates. Let me know if you have any continued problems.

PR #1382, opened 2022-11-07
fix: Register BotWithLookup to fix ArgumentNullException for OnTurn

### LINKED ISSUE #1380: Callback passed to the timer created in TimeoutChoiceInput doesn't work
## Describe the bug
In TimeoutChoiceInput, we create a timer in BeginDialogAsync. We call ContinueConversationAsync as part of this timer by passing in BotWithLookup.OnTurn as the callback. This currently doesn't work since we are not registering BotWithLookup, so OnTurn is null.

## To Reproduce

Add the telephony package to a composer bot, and then add the timeout node to the dialog flow. After the timeout interval, the flow should continue using the default value but nothing happens.

## Expected behavior
The dialog should continue after the timeout interval specified.

ninilang:
Fixes #1380 

### Purpose
This PR registers BotWithLookup as the IBot implementation instead of the AdaptiveDialogBot that comes as default for a composer default. This is because we need to store a reference to the OnTurnAsync method of the AdaptiveDialogBot, so unless we register BotWithLookup, the constructor won't be called and OnTurn will be set to null.

### Tests
No changes in functionality. Fixing a DI issue.

### DIFF
--- a/packages/Telephony/Actions/TimeoutChoiceInput.cs
+++ b/packages/Telephony/Actions/TimeoutChoiceInput.cs
@@ -1,16 +1,13 @@
 ﻿using System;
-using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Security.Claims;
 using System.Threading;
 using System.Threading.Tasks;
 using AdaptiveExpressions.Properties;
 using Microsoft.Bot.Builder;
-using Microsoft.Bot.Builder.Adapters;
 using Microsoft.Bot.Builder.Dialogs;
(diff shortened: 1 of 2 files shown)


stevengum (approved):
LGTM! :shipit: 🎊

stevengum (approved):

stevengum:
Update: @ninilang and I vetted the PR's changes separately and together and have observed it functioning as expected within a single bot experience (i.e., testing the changes by starting a direct conversation with the bot)

Testing with the changes as a skill are still underway

Issue #1378, opened 2022-11-02
[Telephony] "Aggregate Serial Number Input" Preview doesn't terminate nor ignore inputs that definitely wouldn't match


stevengum:
## Describe the bug

The aggregation functionality adds all incoming text to the aggregated inputs stored in state, regardless of if the latest text increases the odds of a match, or worse, if the latest text nullifies any future chance of obtaining a regex match.

This means that if an end-user needed to provide a sequence of 7 numbers, but additional audio input (e.g., background speech or C2 talking to someone else) results in an always-fail state.

|What's occuring|What's sent to bot|Current aggregated inputs| Final aggregate inputs stored|
|-|-|-|-|
| Bot prompts user to provide sequence of 7 numbers | N/A | N/A | N/A|
|User says 1| "1" | ""|"1"|
|User says 7| "7" | "1" |"17"|
|User says "Hello" to someone else| "hello" | "17" |"17hello"|
|User says 4 | "4" | "17hello" | "17hello4" |

As soon as the bot processes "hello" it's impossible to finish the dialog because the logic in the code compares against the aggregate inputs, but it also stores any received input.

So the bot is trying to match `([0-9]{7})` against `"17hello4"`, which will always fail.

## The problematic code
https://github.com/microsoft/botframework-components/blob/eb26e5447d0ff1ce0aab7876fc164a972c2c5daa/packages/Telephony/Actions/SerialNumberInput.cs#L162-L196

## Version
[1.4.0](https://www.nuget.org/packages/Microsoft.Bot.Components.Telephony/1.4.0)

## To Reproduce

Setup bot with a regex pattern (e.g., `([0-9]{7})`) and run it in debug mode, with a breakpoint on the following line

https://github.com/microsoft/botframework-components/blob/eb26e5447d0ff1ce0aab7876fc164a972c2c5daa/packages/Telephony/Actions/SerialNumberInput.cs#L165

Then in the middle of replying to the aggregate serial number input node, say any word and examine the local `existingAggregation` variable. Afterwards, _Continue_ in the debugging process, and observe that no `results` are found, and that `existingAggregation` is stored in state.

## Expected behavior

Invalid inputs aren't stored in state, e.g., no letters are valid when the pattern `([0-9]{7})` is used.
[...]

Issue #1377, opened 2022-11-02
[Telephony] "Aggregate Serial Number Input" Preview node doesn't fully support regex


stevengum:
## Describe the bug

The `Aggregate Serial Number Input` (preview) node doesn't full support regex patterns despite signaling to users that it does. Only `A-Z`, `0-9`, a single character range (`[]`) and single length (`{}`)  are respected. If users provide multiple length requirements, character classes, or other valid regex including meta sequences, alternates, etc, they will not be used at runtime to match aggregate inputs.

Users can provide patterns such as `([a-z0-9]{6})`, `([a-z]{2})`, or `([0-9]{7})` which will work as expected.

This table of example _valid_ Regex patterns and what the TextGroup actually stores as the Regex string:

|Valid Regex|Regex provided by `TextGroup`|Notes|
|--|--|--|
|`([bcBC0-9]{6})`|`([0-9]{6})`| **Upper and lowercase letters B and C are not accepted**|
|`([A-Z1-5]{5})`|`([A-Z]{5})`| **No numbers are accepted**|
|`([\d]{6})`|`([]{6})`| **`\d` is a valid shorthand for `0-9`** |
|`([A-Z]{2}[0-9]{3})`|`([A-Z0-9]{3})`| **"HA121" would be caught by the valid regex, instead "HA", "HA2", "333", "ABC" are all caught by TextGroup's regex and "HA121" would not match** |
|`([A-Z]{2}[0-9]{2,3})`|**N/A**| **This results in a `System.FormatException`**  ([Example](https://dotnetfiddle.net/WhffQf))|

### Where the `TextGroup` determines if letters or numbers are supported, by storing two booleans.
https://github.com/microsoft/botframework-components/blob/eb26e5447d0ff1ce0aab7876fc164a972c2c5daa/packages/Telephony/Actions/SerialNumberTextGroup.cs#L37-L45

### Where the `RegexString` is generated from the `TextGroup`
https://github.com/microsoft/botframework-components/blob/eb26e5447d0ff1ce0aab7876fc164a972c2c5daa/packages/Telephony/Actions/SerialNumberTextGroup.cs#L80-L98

## Version
[1.4.0](https://www.nuget.org/packages/Microsoft.Bot.Components.Telephony/1.4.0)

## To Reproduce

Provide any of the regex included in the table, observe as the bot does not match against them, or throws an exception.

## Expected behavior

Regex patterns to work.

stevengum:
[...]

PR #1389, opened 2022-11-17
Fix aggregation bug in SerialNumberInput and update its configurable properties

stevengum:
Fixes: #1378
Fixes: #1377 by removing the ability for consumers to provide a regex pattern.

## _Important_
The changes in this PR are semantically breaking (i.e., necessitating a major version bump) and the `SerialNumberInput` class may undergo additional changes in future PRs (if not in future iterations of this PR).

While the `SerialNumberInput` class is marked as "preview" when viewing from Composer, it is otherwise not marked as such for code-first consumers of `Microsoft.Bot.Components.Telephony`. Therefore, I consider the PR's changes to be semantically breaking but I'm open to persuasion otherwise.

### Purpose/Changes
- Fixes the aggregation bug of aggregating invalid inputs and entering a permanent failstate
- Removes the ability to provide a regex pattern and simplifies the authoring experience to provide two booleans (`acceptAlphabet` and `acceptNumbers`) and a number (`batchLength`)
- Minor cleanup in `Microsoft.Bot.Components.Telephony`

### Tests
- Added first unit tests for `SerialNumberPattern` and `TextGroup`
- Increased `SerialNumberInput` test coverage by 200%, bringing the grand total to... 6 tests...

### DIFF
--- a/packages/Telephony/Actions/SerialNumberInput.cs
+++ b/packages/Telephony/Actions/SerialNumberInput.cs
@@ -48,10 +48,28 @@ namespace Microsoft.Bot.Components.Telephony.Actions
         public BoolExpression AllowInterruptions { get; set; } = false;
 
         /// <summary>
-        /// Gets or sets the serial number pattern to use to decide when the dialog has aggregated the whole message.
+        /// Gets or sets if SerialNumberInput should accept alphabet characters. 
         /// </summary>
-        [JsonProperty("regexPattern")]
-        public StringExpression RegexPattern { get; set; }
+        /// <value>
+        /// Bool or expression which evalutes to bool.
+        /// </value>
+        [JsonProperty("acceptAlphabet")]
+        public BoolExpression AcceptAlphabet { get; set; }
+
+        /// <summary>
+        /// Gets or sets if SerialNumberInput should accept numbers. 
+        /// </summary>
+        /// <value>
+        /// Bool or expression which evalutes to bool.
+        /// </value>
+        [JsonProperty("acceptNumbers")]
+        public BoolExpression AcceptNumbers { get; set; }
+
+        /// <summary>
+        /// Gets or sets the minimum amount of characters collected before storing the value and ending the dialog.
+        /// </summary>
+        [JsonProperty("batchLength")]
+        public NumberExpression BatchLength { get; set; }
 
         /// <summary>
         /// Gets or sets the property to assign the result to.
@@ -120,9 +138,13 @@ namespace Microsoft.Bot.Components.Telephony.Actions
             }
 
             //Get value of termination string from expression
-            var regexPattern = RegexPattern?.GetValue(dc.State);
-            var snp = new SerialNumberPattern(regexPattern, true);
+            var batchLength = BatchLength.GetValue(dc.State);
+            var acceptAlphabet = AcceptAlphabet.GetValue(dc.State);
+            var acceptNumbers = AcceptNumbers.GetValue(dc.State);
 
+            // TODO: Delete placeholder regex pattern once we remove SerialNumberPattern.
+            var regexPattern = $"([{(acceptAlphabet ? "a-zA-Z" : string.Empty)}{(acceptNumbers ? "0-9" : string.Empty)}]{{{batchLength}}})";
+            var snp = new SerialNumberPattern(regexPattern, true);
             var choices = dc.State.GetValue<string[]>("this.ambiguousChoices");
             var isAmbiguousPrompt = choices != null && choices.Length >= 2;
             if (isAmbiguousPrompt)
@@ -142,7 +164,7 @@ namespace Microsoft.Bot.Components.Telephony.Actions
                         return await PromptUserAsync(dc, cancellationToken).ConfigureAwait(false);
                 }
 
-                if (snp.PatternLength == result.Length)
+                if (batchLength == result.Length)
                 {
                     //If so, save it to the output property and end the dialog
                     //Get property from expression
@@ -157,19 +179,44 @@ namespace Microsoft.Bot.Components.Telephony.Actions
                     return new DialogTurnResult(DialogTurnStatus.Waiting);
                 }
             }
-            
-            //append the message to the aggregation memory state
+
+            //If input is longer than pattern length and wasn't handled as an interruption, silently disregard input.
+            var userInput = dc.Context.Activity.Text;
+
+            if (userInput.Length > batchLength)
+            {
+                return new DialogTurnResult(DialogTurnStatus.Waiting);
+            }
+
+            //Append the message to the aggregation memory state and check if its length is longer than batch length.
             var existingAggregation = dc.State.GetValue(AggregationDialogMemory, () => string.Empty);
-            existingAggregation += dc.Context.Activity.Text;
+
+            //Example happy path: PatternLength is 5, AcceptsAlphabet and AcceptsDigits
+            //If user says "A", "B", "C", "1", "22", do not update state as "ABC122" is invalid. Wait for them to provide valid input.
+            if ((existingAggregation + userInput).Length > batchLength)
+            {
+                return new DialogTurnResult(DialogTurnStatus.Waiting);
+            }
+
+            //TODO: Turn counter? How long is someone in this node?
+            //  Scenarios:
+            //  1) User keeps providing an incorrect serial number (i.e., the aggregated inputs are too long)
+            //  2) User keeps saying "Cancel" on an node configured to be uninterruptable
+            //
+            //  For 1) "Sorry, the provided serial number was ..." + ("Would you like to try again?" || "Would you like to speak to an agent?")
+            //  For 2) When talking to a bot, the user shouldn't enter a stuck state. After a certain number of C1-configurable attempts, the bot should let the user "escape".
+
+            //append the message to the aggregation memory state
+            existingAggregation += userInput;
 
             var results = snp.Inference(existingAggregation);
 
             //Is the current aggregated message the termination string?
             if (results.Length == 1)
             {
+                //If we have a result and its length matches the batch length, set the match on Property and end the dialog.
                 if (snp.PatternLength == results[0].Length)
                 {
-                    //If so, save it to the output property and end the dialog
                     //Get property from expression
                     var property = Property?.GetValue(dc.State);
                     dc.State.SetValue(property, results[0]);
@@ -177,6 +224,7 @@ namespace Microsoft.Bot.Components.Telephony.Actions
                 }
                 else
                 {
+                    //Otherwise update the aggregated inputs and wait for the next input.
                     dc.State.SetValue(AggregationDialogMemory, results[0]);
                     return new DialogTurnResult(DialogTurnStatus.Waiting);
                 }
@@ -190,8 +238,6 @@ namespace Microsoft.Bot.Components.Telephony.Actions
             }
             else
             {
-                //else, save the updated aggregation and end the turn
-                dc.State.SetValue(AggregationDialogMemory, existingAggregation);
                 return new DialogTurnResult(DialogTurnStatus.Waiting);
             }
         }
--- a/packages/Telephony/Actions/SerialNumberInput.cs
+++ b/packages/Telephony/Actions/SerialNumberInput.cs
@@ -35,7 +35,7 @@ namespace Microsoft.Bot.Components.Telephony.Actions
             : base()
         {
             // enable instances of this command as debug break point
-            this.RegisterSourceLocation(sourceFilePath, sourceLineNumber);
+            RegisterSourceLocation(sourceFilePath, sourceLineNumber);
         }
 
         /// <summary>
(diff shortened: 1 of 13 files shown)


packages/Telephony/Actions/SerialNumberInput.cs
```diff
@@ -48,10 +48,28 @@ public SerialNumberInput([CallerFilePath] string sourceFilePath = "", [CallerLin
...
+        public BoolExpression AcceptAlphabet { get; set; }
+
+        /// <summary>
+        /// Gets or sets if SerialNumberInput should accept numbers. 
+        /// </summary>
+        /// <value>
+        /// Bool or expression which evalutes to bool.
+        /// </value>
+        [JsonProperty("acceptNumbers")]
+        public BoolExpression AcceptNumbers { get; set; }
```
stevengum, line 66: I think I'd very much like to remove both of these booleans and have `SerialNumberInput` just accept any alphanumeric inputs, aggregating the inputs until the desired batch length is met.

FYI @jameslew

packages/Telephony/Actions/SerialNumberInput.cs
```diff
@@ -113,23 +131,27 @@ public override async Task<DialogTurnResult> BeginDialogAsync(DialogContext dc,
...
+            var acceptAlphabet = AcceptAlphabet.GetValue(dc.State);
+            var acceptNumbers = AcceptNumbers.GetValue(dc.State);
 
-            string[] choices = dc.State.GetValue<string[]>("this.ambiguousChoices");
-            bool isAmbiguousPrompt = choices != null && choices.Length >= 2;
+            // TODO: Delete placeholder regex pattern once we remove SerialNumberPattern.
+            var regexPattern = $"([{(acceptAlphabet ? "a-zA-Z" : string.Empty)}{(acceptNumbers ? "0-9" : string.Empty)}]{{{batchLength}}})";
+            var snp = new SerialNumberPattern(regexPattern, true);
+            var choices = dc.State.GetValue<string[]>("this.ambiguousChoices");
+            var isAmbiguousPrompt = choices != null && choices.Length >= 2;
```
Jeffders, line 149: >2 [](http://example.com/codeflow?start=73&length=1)

class constant (or configurable property?)

packages/Telephony/Actions/SerialNumberInput.cs
```diff
@@ -113,23 +131,27 @@ public override async Task<DialogTurnResult> BeginDialogAsync(DialogContext dc,
...
+            var batchLength = BatchLength.GetValue(dc.State);
+            var acceptAlphabet = AcceptAlphabet.GetValue(dc.State);
+            var acceptNumbers = AcceptNumbers.GetValue(dc.State);
 
-            string[] choices = dc.State.GetValue<string[]>("this.ambiguousChoices");
-            bool isAmbiguousPrompt = choices != null && choices.Length >= 2;
+            // TODO: Delete placeholder regex pattern once we remove SerialNumberPattern.
+            var regexPattern = $"([{(acceptAlphabet ? "a-zA-Z" : string.Empty)}{(acceptNumbers ? "0-9" : string.Empty)}]{{{batchLength}}})";
+            var snp = new SerialNumberPattern(regexPattern, true);
+            var choices = dc.State.GetValue<string[]>("this.ambiguousChoices");
```
Jeffders, line 148: >choices [](http://example.com/codeflow?start=16&length=7)

Are you sure there are at least 2 of these always? You do an array accessor without checking for null or length...just just want to make sure that's ok.

packages/Telephony/Actions/SerialNumberInput.cs
```diff
@@ -142,56 +164,80 @@ public override async Task<DialogTurnResult> ContinueDialogAsync(DialogContext d
...
             }
-            
-            //append the message to the aggregation memory state
+
+            //If input is longer than pattern length and wasn't handled as an interruption, silently disregard input.
+            var userInput = dc.Context.Activity.Text;
+
+            if (userInput.Length > batchLength)
+            {
+                return new DialogTurnResult(DialogTurnStatus.Waiting);
```
Jeffders, line 188: >ialogTurnResult( [](http://example.com/codeflow?start=28&length=16)

How do they get out of this state? Wouldn't it just wait for more input and always be longer?

Jeffders (approved):
:shipit:

packages/Telephony/Actions/SerialNumberInput.cs
stevengum: Ah I see what happened here, I took away the explicit typing so it looks like a new line, the functionality on the ambiguous checking was otherwise unchanged.

From looking at the code and test case, when we reach this part of the flow, the choices are at least always two because the end user is choosing between one option or another, e.g., "Did you mean A or 8? Press 1 for A and 2 for 8"

packages/Telephony/Actions/SerialNumberInput.cs
```diff
@@ -142,56 +164,80 @@ public override async Task<DialogTurnResult> ContinueDialogAsync(DialogContext d
...
+            //Append the message to the aggregation memory state and check if its length is longer than batch length.
             var existingAggregation = dc.State.GetValue(AggregationDialogMemory, () => string.Empty);
-            existingAggregation += dc.Context.Activity.Text;
 
-            string[] results = snp.Inference(existingAggregation);
+            //Example happy path: PatternLength is 5, AcceptsAlphabet and AcceptsDigits
+            //If user says "A", "B", "C", "1", "22", do not update state as "ABC122" is invalid. Wait for them to provide valid input.
+            if ((existingAggregation + userInput).Length > batchLength)
+            {
+                return new DialogTurnResult(DialogTurnStatus.Waiting);
```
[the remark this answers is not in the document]
stevengum: This early return occurs before we append the `userInput` to the existing aggregation 😄

On line 210 of the latest commit ([a043665](https://github.com/microsoft/botframework-components/pull/1389/files/a043665981ef4117e529a952edb52b59beab7915)), that's where we append the `userInput` to the aggregation, and we only save this state on line 228, where the updated aggregation matches the desired pattern except for length.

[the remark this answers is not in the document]
stevengum: Sorry, I realized I didn't answer your question; we don't need to update the state in these early returns 😄

packages/Telephony/Actions/SerialNumberInput.cs
stevengum: This check is along the lines of `if the user says 233 when the bot is expecting a batch length of 2 (e.g., 23 or 23), wait for the user to say the correct length.`

I'm not entirely happy with the approach I went with here, as it's a very by-the-rails experience happy path.

Ideally, there'd be some logic where C1 can have their bot say "This is what you've currently provided `xyz`" after _n_-amount of attempts to provide what **C2** thinks is correct, but what the bot is not equipped to handle in the current node.

These thoughts are briefly covered in a TODO on lines [201-207](https://github.com/microsoft/botframework-components/pull/1389/files/a043665981ef4117e529a952edb52b59beab7915#diff-5e26f7355c395f29b1b2eaa57b5ec940cbe84ec12647c8e311cd30dde9c20f23R201-R207). I didn't complete exploring that functionality, but I strongly believe that in any voice-first bot, these are scenarios and features that need to be explored and haven't been done so by the code in the SDK or the Telephony extensions.

packages/Telephony/Actions/SerialNumberInput.cs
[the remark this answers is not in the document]
stevengum: Agreed! Since I already changed the explicit typing to be implicit, I'll address this in the next iteration.

packages/Telephony/Actions/SerialNumberInput.cs
```diff
@@ -142,56 +164,80 @@ public override async Task<DialogTurnResult> ContinueDialogAsync(DialogContext d
...
+            }
+
+            //TODO: Turn counter? How long is someone in this node?
```
johnataylor, line 201: why not add the logic now?
you could always compare with a reasonably high constant?

johnataylor (approved):
just one suggestion from me

love the fact that you've aligned this code with the regular style across all the other code

tracyboehrer:
So... There is a failure on Telephony when running the tests.  I haven't looked into enough though.

packages/Telephony/Actions/SerialNumberInput.cs
```diff
@@ -48,10 +48,28 @@ public SerialNumberInput([CallerFilePath] string sourceFilePath = "", [CallerLin
         public BoolExpression AllowInterruptions { get; set; } = false;
 
         /// <summary>
-        /// Gets or sets the serial number pattern to use to decide when the dialog has aggregated the whole message.
+        /// Gets or sets if SerialNumberInput should accept alphabet characters. 
         /// </summary>
-        [JsonProperty("regexPattern")]
-        public StringExpression RegexPattern { get; set; }
```
arturl, line 54: I'm happy you're removing this. Even though this is a source breaking change, I don't think anybody used it, and I cannot imagine that it worked correctly for any but trivial patterns.

packages/Telephony/Actions/SerialNumberInput.cs
arturl: I don't think the user will _ever_ figure out that they said too much and repeat exactly what's needed. At that point they will be most likely shouting "agent, agent!" or just hang up. I would suggest getting C2 out of this state as quickly as possible (either take what you need from the utterance to make the full length and ignore the rest, or just error out) and recommend the bot to always confirm the selection with the option to re-enter as a best practice.

packages/Telephony/Actions/SerialNumberInput.cs
arturl: Some kind of circuit-breaker is needed here. For fixed length input you can count the size of input entered so far and get out if that exceeds the expected length by some number. For pound-terminated sequences you can rely on the user pressing #.

packages/Telephony/Actions/SerialNumberPattern.cs
```diff
@@ -286,18 +285,18 @@ public string[] Inference(string inputString)
...
 
-            int patternIndex = 0;
-            int inputIndex = 0;
+            var patternIndex = 0;
+            var inputIndex = 0;
 
-            List<int> ambiguousInputIndexes = new List<int>();
+            var ambiguousInputIndexes = new List<int>();
             bool isMatch = true;
             string fixedUpString = string.Empty;
```
arturl, line 302: Leaking implementation detail here with `HashSet`.

arturl (approved):
I suggest revisiting the logic where you ignore user input and consider the risk of leaving C2 in a stuck state. Disambiguation logic is in a sorry state and not worth fixing now (I'd like to rip it out completely and re-implement but worry about breaking changes, and whether it's worth it now).

Approved so I don't block you.

packages/Telephony/Actions/SerialNumberPattern.cs
```diff
@@ -390,13 +389,13 @@ private AmbiguousResult CheckAmbiguous(ref int inputIndex)
...
-            char[] delimiters = { ' ', ',', '.', '/', '-' };
+            var result = new AsInResult();
+            var delimiters = new char[] { ' ', ',', '.', '/', '-' };
             var tokens = input.Split(delimiters, StringSplitOptions.RemoveEmptyEntries);
 
             if (tokens.Length > 3 && tokens[0].Length == 1)
             {
-                char proposedChar = '*';
+                var proposedChar = '*';
                 if (tokens[1].ToLowerInvariant() == "as" && tokens[2].ToLowerInvariant() == "in")
```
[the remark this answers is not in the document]
stevengum: Created [[Telephony] Alphanumeric Input code only supports English disambiguation](https://github.com/microsoft/botframework-components/issues/1401) to track this language support issue.

packages/Telephony/Actions/SerialNumberInput.cs
stevengum: After thinking it through, I'm putting together a circuit breaker that is configurable via the `"unexpectedInputsUntilInterruptionsAreAllowed"` property.

It defaults to a value of **2**, and is only respected when `"allowInterruptions"` is **false**.

Here's an example from the docstring for the new property:

    /// E.g. BatchLength is 5, AllowInterruptions is false and UnexpectedInputsUntilInterruptionsAreAllowed is 2.
    /// When a user says "cancel" three times, the third "cancel" will be handled as an interruption.

C2 Users do not need to say the same utterance until it is handled as an interruption.

Using the same configuration above, the user could sequentially say "cancel", "quit", and "exit", and after "quit", "exit" would be handled as an interruption.

Any zero or negative value for `"unexpectedInputsUntilInterruptionsAreAllowed"` will be ignored and the default value of **2** will be used, and _ideally_ a LogWarning would occur. But logging doesn't seem to be a pattern inside of `Microsoft.Bot.Dialogs.Adaptive` or in `Microsoft.Bot.Components.Telephony`, so maybe an error is thrown at runtime 😢 🤷🏼

stevengum:
> So... There is a failure on Telephony when running the tests. I haven't looked into enough though.

Latest commits are passing, but I didn't get a chance to look at the previous build's run...

PR #1394, opened 2022-11-28
fix: Update the current turn count instead of MaxTurnCount in TimeoutChoiceInput

### LINKED ISSUE #1393: [Telephony] Fix default value bug in TimeoutChoiceInput
## Describe the bug

Currently in TimeoutChoiceInput, when ContinueDialogAsync is executed as part of ContinueConversation callback provided to the timer, we are updating this.MaxCount to a value of 1. This is a global property and is affecting future conversations once a timer is executed. This is done so that we don't reprompt the user when we call base.ContinueDialogAsync

We instead need to update this.turnCount, which is local to the action and will not affect any other timeout actions (and conversations as a result).

## Version
1.4.1

## To Reproduce
1. Add a Silence Detection node to the dialog and specify a timeout and a default value. Also specify a value in the "Max turn count" property
2. Let the timer trigger and use the default value for the action
3. Now start a new conversation and trigger the dialog again. When the Silence Detection prompt shows up, send an invalid value to the bot. We can see that we use the default value after the first failed validation instead of reprompting.

## Expected behavior
Once the timer is used, and we start a new conversation with the bot and use the silence detection action again, the MaxTurnCount value should be respected when the user enters an invalid value.

ninilang:
Fixes #1393: Default value bug in TimeoutChoiceInput

### Purpose

We are currently updating the MaxTurnCount instead of this.turnCount for the timer workflow which is affecting other conversations due to MaxTurnCount being a global value.

### Changes

Update this.turnCount instead of MaxTurnCount so the changes are local to the action

### Tests

Added tests to check for the turnCount being updated

### DIFF
--- a/packages/Telephony/Actions/TimeoutChoiceInput.cs
+++ b/packages/Telephony/Actions/TimeoutChoiceInput.cs
@@ -71,7 +71,7 @@ namespace Microsoft.Bot.Components.Telephony.Actions
 
                 //We need to set interrupted here or it will discard the continueconversation event...
                 dc.State.SetValue(TurnPath.Interrupted, true);
-                return await base.ContinueDialogAsync(dc, cancellationToken).ConfigureAwait(false);
+                return await ContinueTimeoutChoiceInputDialogAsync(dc, cancellationToken).ConfigureAwait(false);
             }
             else
             {
@@ -107,6 +107,11 @@ namespace Microsoft.Bot.Components.Telephony.Actions
                 //End dirty hack
             }
 
+            return await ContinueTimeoutChoiceInputDialogAsync(dc, cancellationToken).ConfigureAwait(false);
+        }
+
+        protected virtual async Task<DialogTurnResult> ContinueTimeoutChoiceInputDialogAsync(DialogContext dc, CancellationToken cancellationToken = default(CancellationToken))
+        {
             return await base.ContinueDialogAsync(dc, cancellationToken).ConfigureAwait(false);
         }
 
--- a/packages/Telephony/Actions/TimeoutChoiceInput.cs
+++ b/packages/Telephony/Actions/TimeoutChoiceInput.cs
@@ -64,7 +64,10 @@ namespace Microsoft.Bot.Components.Telephony.Actions
             if (!interrupted && activity.Type != ActivityTypes.Message && activity.Name == ActivityEventNames.ContinueConversation)
             {
                 //Set max turns so that we evaluate the default when we visit the inputdialog.
-                MaxTurnCount = 1;
+                if (MaxTurnCount != null)
+                {
+                    dc.State.SetValue(TURN_COUNT_PROPERTY, this.MaxTurnCount.GetValue(dc.State) + 1);
+                }
 
                 //We need to set interrupted here or it will discard the continueconversation event...
                 dc.State.SetValue(TurnPath.Interrupted, true);
--- a/packages/Telephony/Actions/TimeoutChoiceInput.cs
+++ b/packages/Telephony/Actions/TimeoutChoiceInput.cs
@@ -1,4 +1,7 @@
-﻿using System;
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
 using System.Runtime.CompilerServices;
 using System.Security.Claims;
 using System.Threading;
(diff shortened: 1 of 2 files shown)


Jeffders:
>                 dc.State.SetValue(VALUE_PROPERTY, null);

Just checking, but you do want the Continue called for the "else" case above?

---
Refers to: packages/Telephony/Actions/TimeoutChoiceInput.cs:105 in 7a1de66. [](commit_id = 7a1de66e47d1721bc250fe29550bbf0e42778976, deletion_comment = False)

Jeffders (approved):
:shipit:

ninilang:
@Jeffders Yes the Continue is needed for the else part. The else part is just to start a new timer in case the user entered option is incorrect.

tests/unit/packages/Microsoft.Bot.Components.Telephony.Tests/TimeoutChoiceInputTests.cs
```diff
@@ -0,0 +1,67 @@
...
+
+namespace Microsoft.Bot.Components.Telephony.Tests
+{
+    public class TimeoutChoiceInputTests
+    {
+        [Fact]
+        public async Task TimeoutChoiceInput_TurnCountIsCorrectlyUpdated()
+        {
+            // Setup
+            Mock<ITurnContext> mockTurnContext = new Mock<ITurnContext>();
```
stevengum, line 21: Could you please you use `var` for these `new`-ed up instances to be more consistent and brief in the code base? 😃 

I see additional explicit typing for `DialogStateManagerConfiguration`, `DialogContext`, and `TimeoutChoiceInputTestHelper`; where `var` could be used instead.

packages/Telephony/Actions/TimeoutChoiceInput.cs
```diff
@@ -104,6 +107,11 @@ public override async Task<DialogTurnResult> ContinueDialogAsync(DialogContext d
                 //End dirty hack
             }
 
+            return await ContinueTimeoutChoiceInputDialogAsync(dc, cancellationToken).ConfigureAwait(false);
+        }
+
+        protected virtual async Task<DialogTurnResult> ContinueTimeoutChoiceInputDialogAsync(DialogContext dc, CancellationToken cancellationToken = default(CancellationToken))
```
stevengum, line 113: I believe the code here can be made terser by removing `CancellationToken)`

```suggestion
        protected virtual async Task<DialogTurnResult> ContinueTimeoutChoiceInputDialogAsync(DialogContext dc, CancellationToken cancellationToken = default)
```

tests/unit/packages/Microsoft.Bot.Components.Telephony.Tests/TimeoutChoiceInputTests.cs
```diff
@@ -0,0 +1,67 @@
...
+            // Assert
+            Assert.Equal(4, dc.State.GetValue<int>("this.turnCount", () => 0));
+            Assert.True(dc.State.GetValue<bool>(TurnPath.Interrupted, () => false));
+        }
+
+        private class TimeoutChoiceInputTestHelper : TimeoutChoiceInput
+        {
+            public TimeoutChoiceInputTestHelper(): base() { }
+
+            protected override Task<DialogTurnResult> ContinueTimeoutChoiceInputDialogAsync(DialogContext dc, CancellationToken cancellationToken = default(CancellationToken))
```
stevengum, line 61:
```suggestion
            protected override Task<DialogTurnResult> ContinueTimeoutChoiceInputDialogAsync(DialogContext dc, CancellationToken cancellationToken = default)
```

packages/Telephony/Actions/TimeoutChoiceInput.cs
```diff
@@ -64,11 +64,14 @@ public override async Task<DialogTurnResult> ContinueDialogAsync(DialogContext d
             if (!interrupted && activity.Type != ActivityTypes.Message && activity.Name == ActivityEventNames.ContinueConversation)
             {
                 //Set max turns so that we evaluate the default when we visit the inputdialog.
-                MaxTurnCount = 1;
+                if (MaxTurnCount != null)
+                {
+                    dc.State.SetValue(TURN_COUNT_PROPERTY, this.MaxTurnCount.GetValue(dc.State) + 1);
+                }
```
stevengum, line 70: Should this comment be deleted since the functionality and comment were incorrect?

ninilang: No this comment is still relevant. We are just setting max turns for the action in a different way now

stevengum (approved):
:shipit: Thanks for addressing the feedback!

Issue #1551, opened 2024-01-31
Azure AI Language Orchestration Workflow Recognizer


RadoslavGatev:
## Describe the solution you'd like

A recognizer that integrates with the AI Language Orchestration Workflow and is capable of switching between Conversational Language Understanding and Custom Question Answering intents.

## Describe alternatives you've considered

N/A

## Additional context
N/A

PR #1552, opened 2024-01-31
feat: Azure AI Language Orchestration Workflow Recognizer

RadoslavGatev:
### Purpose

This is a proposal for the implementation of #1551. 

### Changes

N/A

### Tests

At this point only manual tests  were done in Composer.


RadoslavGatev:
The majority of the code is based on the existing CLU recognizer and also uses some concepts from Custom Q&A recognizer. The code is not polished yet!

If you are willing to accept this PR, there are two options I can think of:

1. Leave it as a standalone recognizer named `Microsoft.OrchestrationRecognizer`
2. Merge it with `Microsoft.CluRecognizer`. But then the name of `Microsoft.CluRecognizer` likely needs to be changed to something else as it hosts both CLU and Orchestration Workflow type of projects.

